PR #19, opened 2019-06-04
Add field button part functionality

NeilAllavarpu:
The add field button now should add the appropriate field to the appropriate part of the form, and be able to handle conditionals. More work is needed to refine it further.


dhuangg:
looks good overall. I'll close this pull request - we need to fully finish this feature before we can merge this.

PR #20, opened 2019-06-05
Form functionality

NeilAllavarpu:
More refinement regarding the forms and the add field buttons.


omou-frontend/src/reducers/initialState.js
```diff
@@ -1,3 +1,481 @@
...
+                        name: "Zip Code",
+                        type: "string",
+                    },
+                ],
+            },
+            tutoring: {
+                form_type: "tutoring",
+                section_titles: ["Tutoring Session Type", "Student(s)", "Student Information", "Tutor Selection", "Payment"],
+                field_limits: {
+                    "Select tutoring type": 1,
```
dhuangg: instead of having field_limits as its own object, just have each field object have a `field_limits` property. For example: 
```
{
    field: "City",
    name: "City",
    type: "short text",
    field_limit: 1,
},
```

dhuangg:
@NeilAllavarpu 
**Background**
Every input field has a value attribute. For example, `<Select value={this.state.selectValue}/>` As you can tell, the value of the form is being stored in the state (because this info should be component based).

 If you look at `componentWillMount()` for Form.js (line 44), you'll notice that I'm creating a value variable associated with each field taken from the Redux state within the Form.js state (this is done by section). But you'll notice on line 62, that it's only creating this field variable for sections where fields are listed in an array (not an object) - this is to support conditional fields as you may remember. 
Screen Shot 2019-06-05 at 2 30 07 PM

Screen Shot 2019-06-05 at 2 24 24 PM

**What you'll need to do is**: create the field value variable in state for each field (i.e. Student Names) even for conditional fields. When you do this, you will also have to create a `_validated` variable in state for each field in the conditional section.

dhuangg:
@NeilAllavarpu second issue - right now each field that can have additional fields has a "Add [Field]" button. If they're the same type of field, we only want the button to show up once. 

You'll also notice these fields are invalid at the start - that's because the `Student(s)_validated` section in `this.state` is only an empty object with no values (reference console.log screenshot from the previous comment). You'll need to fill this section up with something. 

Screen Shot 2019-06-05 at 2 33 44 PM

PR #21, opened 2019-06-07
Autocomplete for course registration when clicked from the main registration menu

NeilAllavarpu:
Parses the parameters in the URL with encode/decodeURIComponent and then appropriately fills out the selected course when a course field first appears.


omou-frontend/src/components/FeatureViews/Registration/Form.js
```diff
@@ -150,31 +142,45 @@ class Form extends Component {
...
             } else {
                 return {};
             }
+        }, () => {
+            const field = this.getActiveSection().find((matchingField) => matchingField.type === "course");
+            if (field) {
+                let course = decodeURIComponent(this.props.match.params.course);
+                course = this.props.courses.find((matchingCourse) => course === matchingCourse.course_title);
+                // convert it to a format that onselectChange can use
+                course = {
```
dhuangg: The logic of when to setting the value for courses shouldn't be in `handleNext()` in the case that the course is the first section (which is highly likely - we want to put course selection as the first section)

PR #37, opened 2019-06-19
Improved form functionality and validation

### LINKED ISSUE #28: Registration checks
There are a couple of checks that we need to perform on the form

Course Registration
- [x] If the course select has room for registration => if not, remove from list of possible options

Create New Student
- [x] if student already exists (check against email) => if student exists, display an error
- [x] if parent already exists (check against email) => if parent already exists, autofill parent info

NeilAllavarpu:
Close #28

Validation should function properly for all fields now, buttons are disabled when their action is invalid, form input is preserved even if the page is reloaded, and select field functionality has been improved.

### DIFF
--- a/omou-frontend/src/components/FeatureViews/Registration/Form.js
+++ b/omou-frontend/src/components/FeatureViews/Registration/Form.js
@@ -67,10 +67,16 @@ class Form extends Component {
                         NewState[title] = {};
                         // set a value for every non-conditional field (object)
                         if (Array.isArray(formContents[title])) {
-                            formContents[title].forEach((field) => {
-                                NewState[title][field.field] = undefined;
-                                if (field.type === "course") {
-                                    NewState[title][field.field] = course;
+                            formContents[title].forEach(({field, type, options}) => {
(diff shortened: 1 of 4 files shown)


omou-frontend/src/components/FeatureViews/Registration/Form.js
```diff
@@ -210,74 +210,82 @@ class Form extends Component {
...
         switch(field.type){
             case "select":
-                console.log(this.state, label);
                 return <FormControl className={"form-control"}>
                     <InputLabel htmlFor={fieldTitle}>{fieldTitle}</InputLabel>
                     <Select
-                        value={this.state[label][fieldTitle] ? this.state[label][fieldTitle] : "Private Tutoring"}
-                        onChange={(e)=>{this.onSelectChange.bind(this)(e.target.value,label,fieldTitle)}}
-                    >
+                        value={this.state[label][fieldTitle]}
```
dhuangg: make default value to be "Private Tutoring" need to update the state for this

PR #40, opened 2019-06-22
I made a Test comment

ryanliou02:


omou-frontend/src/index.js
```diff
@@ -18,6 +18,7 @@ ReactDOM.render(
     document.getElementById('root')
 );
 
+// test ryan
```
dhuangg: this is great! jk delete it

PR #42, opened 2019-06-22
Back Button Component

NeilAllavarpu:
Moved the back button to its own component. Currently it can be configured to offer a warning on exit or not, as well as run a function when the page goes back.

PR #46, opened 2019-06-26
Revert "Back Button Component"

jsflca2012:
Reverts omou-org/front-end#42


jsflca2012:
I accidentally committed the pull request for #42

dhuangg:
@jsflca2012 this reverts it back to before there was a merge button so we don't want this. I fixed the issues i had with the original pull request you made/approved. My comments on your original pull request: Be sure to do more testing to see if there are stray console logs (there was one in Form.js), if the function name is `this.denyAction` it's fairly clear what function it calls so you don't need to comment `//calls deny action`, in general just remember I should be the only one merging pull requests

PR #65, opened 2019-07-14
needs fixing

jsflca2012:
Delete button stopped working.


dhuangg:
you need to be more descriptive. what isn't working? why do you think it's
not working?

dhuangg:
@jsflca2012  this function doesn't behave the way you want it to (i.e. getting the list of fields from the initial state). Tell me what does it do:
```
function getSectionFieldList(path, formList) {
    if (Array.isArray(path)) {
        if (path.length === 0) {
            return formList;
        }
        return getSectionFieldList(path, formList[path.shift()])
    }
    Error("Path variable not an array");
}
```

PR #67, opened 2019-07-22
Added teacher button

CalvinFronda:
Created a button on Registration section that leads to "Add Teacher" form.

Changed UI features to make Gender & State inputs a selection.


omou-frontend/src/components/Form/Form.js
```diff
@@ -307,24 +308,24 @@ class Form extends Component {
                     };
                     console.log(NewState[label]);
```
dhuangg: this may have been leftover from something we did before. can you delete this?

omou-frontend/src/components/Form/Form.js
```diff
@@ -495,7 +501,8 @@ class Form extends Component {
             default:
                 return <TextField
                     label={field.field}
-                    multiline
+                    //Sets multiline to false for default forms
+                    multiline={this.setState.multiline === false}
                     // className={this.state[label+"_validated"][field.field] ? "": "error"}
```
dhuangg: Did you comment this out?

CalvinFronda: I added the setState below that line

omou-frontend/src/components/Form/Form.js
CalvinFronda: Got it

omou-frontend/src/components/Form/Form.js
```diff
@@ -495,7 +501,8 @@ class Form extends Component {
             default:
                 return <TextField
                     label={field.field}
-                    multiline
+                    //Sets multiline to false for default forms
+                    multiline={this.setState.multiline === false}
```
dhuangg: Each field needs to be defined as multiline or not. you'll need to give each field from `initialState.js` a new "multiline" property with a true/false value. right now the entire form would not be multiline. 

For example, on `initialState.js` line 826
```
{
                        field: "Background",
                        name: "Background",
                        type: "string",
                        full: true,
                        required: false,
                        field_limit: 1,
                    },
```
there's no "multiline" property. we want to add it here so it looks like: 
```
{
                        field: "Background",
                        name: "Background",
                        type: "string",
                        full: true,
                        required: false,
                        field_limit: 1,
                        multiline: true, <-new prop
                    },
```
so when you're in `Form.js` within the TextField tag (line 505), you can write the code as: `multiline = { field.multiline }`

CalvinFronda: Got it, thanks!

PR #78, opened 2019-08-21
Session view

CalvinFronda:


omou-frontend/src/components/FeatureViews/Scheduler/SessionView.js
```diff
@@ -0,0 +1,185 @@
...
+        };
+
+
+
+        return (
+            <Grid className="">
+                <Paper className={"paper"}>
+                    <SessionActions />
+                </Paper>
+                <Grid className={'main-session-container'}>
```
dhuangg: is this Grid an item or container? you'll need to specify as `<Grid container className={'asdf'}/>

omou-frontend/src/components/FeatureViews/Scheduler/SessionView.js
```diff
@@ -0,0 +1,185 @@
...
+                        <Grid item xs={12}>
+                            <Typography className={'session-view-title'} align={'left'} variant={'h3'}>
+                                {this.state.courseData.title}
+                            </Typography>
+                        </Grid>
+                        <List style={flexContainer}>
+                            <ListItem>
+                                <ListItemText
+                                    primary="Subject"
+                                    secondary={this.state.courseData.subject} />
```
dhuangg: What is secondary text used for? if it's what I think it is, you should use the <Typography> component where you can specify text styles. for example:  
```
<Grid item md={12}>
    <Typography variant = "h2"> Subject </Typography>
    <Typography varient = "p">{this.state.courseData.subject} </Typography>
</Grid>
```

CalvinFronda: Secondary texted is used for text below the primary.
So in any instance it will say the subject of the session like 
Subject
Math

You don't need to mess with any css to get the words aligned. 
If you go to https://material-ui.com/components/lists/#lists > interactive. You can see how the secondary works.

omou-frontend/src/components/FeatureViews/Scheduler/SessionView.js
CalvinFronda: it should be a container, I'll fix right now

omou-frontend/src/components/FeatureViews/Scheduler/SessionView.js
CalvinFronda: Actually you way seems easier to style the words. I'll make the changes.

PR #99, opened 2019-09-11
Test scripts

CalvinFronda:


omou-frontend/src/App.test.js
```diff
@@ -1,9 +1,78 @@
...
-
-it('renders without crashing', () => {
-  const div = document.createElement('div');
-  ReactDOM.render(<App />, div);
-  ReactDOM.unmountComponentAtNode(div);
+const puppeteer = require('puppeteer');
+const faker = require('faker');
+
+const person = {
+  name: faker.name.firstName() + ' ' + faker.name.lastName(),
```
NeilAllavarpu: Try to use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) to have concatenated strings, this generally makes the structure of the string clearer.

omou-frontend/src/App.test.js
```diff
@@ -1,9 +1,78 @@
...
+      height: 768
+    },
+    userAgent: ''
+  });
+
 });
+
+
+
+const appUrlBase = 'http://localhost:3000'
```
NeilAllavarpu: Move this to the top of the file so that all functions can use it.

omou-frontend/src/App.test.js
```diff
@@ -1,9 +1,78 @@
...
+  };
+  return process.env.NODE_ENV === 'debug' ? debugging_mode : {};
+};
+
+let browser;
+let page;
+beforeAll(async () => {
+  browser = await puppeteer.launch(isDebugging());
+  page = await browser.newPage();
+  await page.goto('http://localhost:3000/');
```
NeilAllavarpu: Use the constant `appUrlBase` instead of hardcoding the URL, so it can be easily changed later if need be.

omou-frontend/src/App.test.js
```diff
@@ -1,9 +1,78 @@
...
+    },
+    userAgent: ''
+  });
+
 });
+
+
+
+const appUrlBase = 'http://localhost:3000'
+const routes = {
```
NeilAllavarpu: Also move this up.

omou-frontend/src/App.test.js
```diff
@@ -1,9 +1,78 @@
...
+  private: {
+    dashboard: `${appUrlBase}/`,
+    registration: `${appUrlBase}/registration`,
+  },
+};
+
+describe('Signing In', () => {
+  test('sign-in title loads correctly', async () => {
+
+    await page.goto('http://localhost:3000/');
```
NeilAllavarpu: Did you mean to use the login page? If so, you can use `routes.public.login` instead of hardcoding the url.

omou-frontend/src/App.test.js
```diff
@@ -1,9 +1,78 @@
...
+
+    await page.click('.email')
+    await page.type('.email', person.email)
+    await page.click('.password')
+    await page.type('.password', person.password)
+    await page.click('.remember')
+    await page.click('.signIn')
+
+    const html = await page.$eval('.Navigation', e => e.innerHTML);
+    expect(html).toBe('.Navigation')
```
NeilAllavarpu: If you look at the test, the `innerHTML` is much different from `.Navigation`; did you mean to compare the classes of the found objects? Check available properties at [MDN](https://developer.mozilla.org/en-US/docs/Web/API/element).

PR #139, opened 2019-10-19
Test-Script boiler

CalvinFronda:
updated test scripts to work with new development


src/App.test.js
```diff
@@ -1,9 +1,87 @@
...
+    await page.type('.email', person.email)
+    await page.click('.password')
+    await page.type('.password', person.password)
+    await page.click('.remember')
+    await page.click('.signIn')
+
+    //This is supposed to show Dashboard, but can't find selector of `Dashboard`
+    const html = await page.$eval('`Dashboard`', e => e.innerHTML);
+
+    expect(html).toBe('Dashboard')
```
NeilAllavarpu: You are trying to check if the Dashboard has rendered (you needed a `.` in front of the class name to select for a class), but a simpler way to check if login was successful is to see if the [url](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pageurl) has changed. You'll need to [wait](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagewaitforselectororfunctionortimeout-options-args) for a bit to ensure the request has time to go through to the backend (500 milliseconds seems to work), and then check to see if the URL matches the dashboard's url.

src/App.test.js
```diff
@@ -1,9 +1,87 @@
...
+
+let browser;
+let page;
+const width = 1920;
+const height = 1080;
+
+beforeAll(async () => {
+  browser = await puppeteer.launch({
+    headless: false,
+    slowMo: 80,
```
NeilAllavarpu: This should probably default to having no `slowMo` or browser opening to run tests faster (you shouldn't need to see exactly what is going on for other people's component tests that you didn't touch, just to see if it didn't break - if it does break, then you can turn this back on to see what happened).

src/App.test.js
```diff
@@ -1,9 +1,87 @@
...
+    await page.waitForSelector('.header');
+
+    const html = await page.$eval('.header', e => e.innerHTML);
+    expect(html).toBe('sign in');
+
+
+  }, 16000);
+  test('login form works correctly', async () => {
+    // take a png screenshot and saves it inside root directory
+    await page.screenshot({ path: "loginScreen.png" })
```
NeilAllavarpu: Save screenshots to a new folder (e.g. `debug_screenshots` or similar) to avoid cluttering the root folder.

src/App.test.js
```diff
@@ -1,9 +1,87 @@
...
+  }, 16000)
+}
+);
+
+describe('Register', () => {
+  // xtest means skip this test
+  xtest('Go to register page click on tutoring and click 6th grade math ', async () => {
+    await page.goto(routes.private.registration);
+    await page.click('a[href$="/registration/form/course/1"]')
+    await page.screenshot({ png: "register.png" })
```
NeilAllavarpu: Did you mean to type `path` here?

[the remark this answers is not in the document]
CalvinFronda: sounds good, but it needs to be the exact path. I've tired to send it to 
`await page.screenshot({ path : "../debug_screenshot/test.png })` 
but this won't work. 

This works!
`await page.screenshot({ path: "/Users/calvinfronda/Documents/GitHub/front-end/debug_screenshots/test.png" })`

src/App.test.js
```diff
@@ -1,9 +1,87 @@
...
-
-it('renders without crashing', () => {
-  const div = document.createElement('div');
-  ReactDOM.render(<App />, div);
-  ReactDOM.unmountComponentAtNode(div);
+const puppeteer = require('puppeteer');
+const faker = require('faker');
+
+const person = {
+  name: faker.name.firstName() + ' ' + faker.name.lastName(),
```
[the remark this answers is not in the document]
CalvinFronda: :+1: you got it boss

src/App.test.js
CalvinFronda: if you don't need to browser to open you can just set `headless: true`. slowMo is slowing the speed of puppeteer by millisecond. I've only had it on for the login demonstration.

src/App.test.js
CalvinFronda: Tired to set the dashboard class with  `.` but got "not found error". If you look at the dashboard it has a class of `<div className="``Dashboard``">` with the backticks. 

I've changed the test per your request to be 
`expect(window.location.pathname).toBe('/')`

src/App.test.js
```diff
@@ -54,10 +57,12 @@ describe('Signing In', () => {
...
 
   }, 16000);
-  test('login form works correctly', async () => {
-    // take a png screenshot and saves it inside root directory
-    await page.screenshot({ path: "loginScreen.png" })
-    // uses faker to load data from object uptop
+  test('login form leads to the dashboard', async () => {
+
+    // you will need to find the exact path for the screen shots to save in the file. 
+    // i've tried ({path : "../debug_screenshot"}) but you will get a Unsupported screenshot mime type: null error
```
NeilAllavarpu: `path: "debug_screenshot/login.png"` worked for me (you may have forgotten the extension).

CalvinFronda: you right

src/App.test.js
```diff
@@ -1,9 +1,90 @@
...
+let browser;
+let page;
+const width = 1920;
+const height = 1080;
+
+// Setup before any test is run, launches puppeteer with the window sizes & heigh
+beforeAll(async () => {
+  browser = await puppeteer.launch({
+    headless: true,
+    slowMo: 80,
```
NeilAllavarpu: Remove `slowMo` (just to make it a bit faster when testing), and add a `page.wait()` at necessary parts (e.g. when an API request is made, when something new needs to load, etc.) If this turns out to be unwieldy, we can revert this later.

src/App.test.js
```diff
@@ -1,9 +1,90 @@
...
+    await page.screenshot({ path: "debug_screenshots/test.png" })
+    // Uses faker to load data from object uptop
+    await page.goto(routes.public.login);
+    await page.click('.email')
+    await page.type('.email', person.email)
+    await page.click('.password')
+    await page.type('.password', person.password)
+    await page.click('.remember')
+    await page.click('.signIn')
+
```
NeilAllavarpu: This would most likely need a `page.wait()` when `slowMo` is removed.

CalvinFronda: I took out `slowMo` and the test finish in 5 secs. Where do i add page.wait()

NeilAllavarpu: After you click sign in, but before you check the URL (give time for login to be validated and the redirect to occur).

src/App.test.js
```diff
@@ -70,6 +69,7 @@ describe('Signing In', () => {
     await page.type('.password', person.password)
     await page.click('.remember')
     await page.click('.signIn')
+    await page.waitFor(".Navigation")
```
NeilAllavarpu: Err, try waiting for a set time, because if this happens to redirect to something aside from the Dashboard, where there is no Navigation-class element, this would stall (as would be the case currently on stage).

PR #235, opened 2020-01-29
OOO

jsflca2012:
Currently confused as to why this.setstate is setting open into an infinite loop, need advice on how it should be done.


src/components/FeatureViews/Accounts/ProfileHeading.js
```diff
@@ -231,6 +253,13 @@ class ProfileHeading extends Component {
                         {this.props.isAdmin && this.renderEditButton()}
                     </Grid>
                 </Grid>
+                <Grid>
+                    <Button onClick={this.handleClick()}>
```
dhuangg: The onClick event handler's default action is to [automatically trigger this handler](https://javascript.info/default-browser-action) whether or not you've actually clicked. Solve this issue by looking at my comment above.

src/components/FeatureViews/Accounts/ProfileHeading.js
```diff
@@ -190,9 +203,16 @@ class ProfileHeading extends Component {
                 </>
             );
         }
+
     }
 
+handleClick(){
```
dhuangg: add "event" in the parenthesis

src/components/FeatureViews/Accounts/ProfileHeading.js
```diff
@@ -190,9 +203,16 @@ class ProfileHeading extends Component {
...
+    this.setState({open:true,})
```
dhuangg: Add a new line before this line with this line:
```
event.preventDefault();
```

jsflca2012: binding the button to onclick={this.handleClick().bind(this)}sets all buttons to the same function, is there a way for me to bind it to only one button?

dhuangg: Rename your function? I'm not quite certain what you're asking specifically. 

I also just realized this, but you're not using a react hook. React Hooks will make your life significantly easier. Here are some potential references:
- [Egghead tutorial on refactoring react component class to hook](https://egghead.io/lessons/react-refactor-a-class-component-with-react-hooks-to-a-function)
- [Blog post on the same topic](https://levelup.gitconnected.com/refactoring-react-class-components-with-hooks-d15abb7fe71a)

src/components/FeatureViews/Accounts/ProfileHeading.js
NeilAllavarpu: There should not be parentheses here - otherwise the function is called every render, instead of onClick

src/components/FeatureViews/Accounts/ProfileHeading.js
jsflca2012: I was trying to use hooks, I found the problem was the function was not capitalized, and therefore not rendering as a hook.

jsflca2012: I'm having some trouble getting the dialog to close, I've pushed up the latest updates I've made so you can take a look.

src/components/FeatureViews/Accounts/outOfOffice.js
```diff
@@ -1,35 +1,115 @@
...
-class outOfOffice extends React.Component {
-    constructor(props) {
-        super(props);
-        this.state = {
-        };
-    }
+function OutOfOffice(props) {
+    console.log(props);
+    const setDialog = useState(props.open);
+    const setDate = useState(null);
```
NeilAllavarpu: [setState](https://reactjs.org/docs/react-component.html#setstate) syntax is like this `const [value, setValue] = useState(initialValue);`

src/components/FeatureViews/Accounts/ProfileHeading.js
jsflca2012: Error is this: 
image

dhuangg:
@jsflca2012 you're not using the useState() function correctly. Refer to
Neil's example, you're missing something on line 14 in your array.

src/components/FeatureViews/Accounts/ProfileHeading.js
NeilAllavarpu: You need to declare state variables as:

    const [dialog, setDialog] = useState(props.open);
    const [date, setDate] = useState(null);

[useState](https://reactjs.org/docs/react-component.html#setstate) returns `[value, setter]`, where `value` is the current value of the state variable, and `setter` is how you set the state variable.

jsflca2012:
Having some trouble overriding the default dialog paper class, currently throwing a TypeError: Object(...) is not a function

NeilAllavarpu:
Did you mean to use `className` instead of `classes`?

jsflca2012:
No, that was intentional, I was looking at https://stackoverflow.com/questions/47698037/how-can-i-set-a-height-to-a-dialog-in-material-ui

dhuangg:
@jsflca2012 if you're ever running into an error, be sure to post a screenshot of the error + the lines of code you think are relevant

NeilAllavarpu:
Try the `withStyles(styles)(YourDialog);` wrapper around your `export` at the bottom, seen at the bottom of the linked code.

jsflca2012:
Doing so didn't fix the problem.
For reference, here's a screenshot: 
image

jsflca2012:
Ready for functionality to be implemented @NeilAllavarpu

dhuangg:
@NeilAllavarpu @jsflca2012 you need to display the OOO.....?

Issue #220, opened 2020-01-19
Add "Session Duration" field to Session View


dhuangg:
## Background
Previously there were issues with updating session duration due to how we were handling generating sessions. However, by generating 30 sessions for all courses, we will be calculating the payment status of each session by deducting previous session tuitions from the total enrollment balance. 

## Development
- [ ] add a "Session Duration" field to the edit session view
- [ ] Calculate + PATCH the new end time to the session/course depending on what the user wants to edit.

dhuangg:
@CalvinFronda I'll handle this, but I want you to perform the review for the pull request to understand what/how I'm doing this.

dhuangg:
## Development Todo's 
Everything will be in EditSessionView.js
- [ ] Create a Select field for duration. The option labels the user should select are: "1 Hour" "1.5 Hours" "2 Hours" and "0.5 Hour". The option values should be doubles: 1.0, 1.5, 2.0, 0.5 
- [ ] Create a `handleDurationSelect()` function where depending on which option the user select, you update the "end_time" in sessionFields (refer to line 35) 

Note: To figure out the end_time, add the duration time to the start time. Do the math here, and just confirm that you're getting the correct result. Do some Googling.

PR #244, opened 2020-02-04
#220 session duration

CalvinFronda:


src/components/FeatureViews/Scheduler/EditSessionView.js
```diff
@@ -40,6 +40,7 @@ function EditSessionView({ course, session, editSelection }) {
         category: "",
         is_confirmed: "",
         duration: "",
+        tempDuration: ""
```
dhuangg: isn't this just the end time?

src/components/FeatureViews/Scheduler/EditSessionView.js
```diff
@@ -139,19 +136,19 @@ function EditSessionView({ course, session, editSelection }) {
         setSessionFields({
             ...sessionFields,
             duration: event.target.value,
-
+            tempDuration: newEndTime
```
dhuangg: update "end_time" not this "tempDuration", this is too confusing

src/components/FeatureViews/Scheduler/EditSessionView.js
```diff
@@ -235,11 +247,34 @@ function EditSessionView({course, session, editSelection}) {
...
+                            })}
+                        </Select>
+                        {/* <SearchSelect
+                            className="search-options"
+                            isClearable
+                            onChange={handleDurationSelect}
+                            placeholder={"Choose a Duration"}
+                            value={sessionFields.duration}
+                            options={courseDurationOptions}
+                        /> */}
```
dhuangg: Delete the comment

src/components/FeatureViews/Scheduler/EditSessionView.js
```diff
@@ -235,11 +247,34 @@ function EditSessionView({course, session, editSelection}) {
...
+                    <Grid item xs={6}>
+                        <Typography variant="h5"> Duration </Typography>
+                        <Select
+                            onChange={handleDurationSelect}
+                            value={sessionFields.duration}
+                        >
+                            {courseDurationOptions.map((duration) => {
+                                return <MenuItem
+                                    value={duration}
+                                >{duration + " hour(s)"}
```
dhuangg: Indent + add a new line after the >

src/components/FeatureViews/Scheduler/EditSessionView.js
```diff
@@ -235,11 +247,34 @@ function EditSessionView({course, session, editSelection}) {
...
 
+                    <Grid item xs={6}>
+                        <Typography variant="h5"> Duration </Typography>
+                        <Select
+                            onChange={handleDurationSelect}
+                            value={sessionFields.duration}
+                        >
+                            {courseDurationOptions.map((duration) => {
+                                return <MenuItem
+                                    value={duration}
```
dhuangg: add a "key" attribute. Anytime you're using ".map()" and returning a component, do this. This is to improve performance. 

example:
```
courseDurationOptions.map((duration, index) => <MenuItem key={ index } value={ duration }> {duration </MenuItem>
```

src/components/FeatureViews/Scheduler/EditSessionView.js
CalvinFronda: No, this is because when a user chooses an option, it will create the end_time. But I thought of a case where a user made a mistake and needed to change the duration. That would add to the current end time because of how the switch statement is created. So I created a tempDuration to hold the string until a user submits the form and the final temp duration will become the end_time. If you have another more eloquent way of coding this lmk.

src/components/FeatureViews/Scheduler/EditSessionView.js
CalvinFronda: see above

src/components/FeatureViews/Scheduler/EditSessionView.js
dhuangg: I see, that makes sense. The code is fine, it's the naming. Call it "updatedDuration" then. "tempDuration" doesn't mean anything.

Issue #185, opened 2019-12-21
Instructor Select should use Search endpoint


NeilAllavarpu:
Instead of GETting all instructors, use provided input to show a filtered instructor list.

PR #249, opened 2020-02-09
Async Instructor Select

NeilAllavarpu:
Addresses #185

### DIFF
--- a/src/components/Form/Form.js
+++ b/src/components/Form/Form.js
@@ -9,6 +9,8 @@ import React, {Component} from "react";
 import {Prompt} from "react-router";
 import {NavLink, withRouter} from "react-router-dom";
 import CreatableSelect from 'react-select/creatable';
+import * as searchActions from "actions/searchActions";
+import AsyncSelect from "react-select/async";
 import {updateParent, updateStudent} from "reducers/usersReducer";
 import {updateCourse} from "reducers/courseReducer";
 // Material UI Imports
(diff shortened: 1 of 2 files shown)


src/components/Form/Form.js
```diff
@@ -906,24 +910,48 @@ class Form extends Component {
...
-                </div>);
+                const mapInstructor = (id, name, email) => ({
+                    "value": id,
+                    "label": `${name} - ${email}`,
+                })
+
+                instructorList = Object.values(instructorList)
+                    .map(({user_id, name, email}) => mapInstructor(user_id, name, email));
+
+                const loadInstructors = async (inputValue) => {
```
dhuangg: Let's move the async function out of Form.js since Form.js is really unweildy already. Move it to FormUtils.js

src/components/Form/Form.js
```diff
@@ -906,24 +910,48 @@ class Form extends Component {
...
+                    }
+                    return response.data
+                        .map(({"user": {user_id, first_name, last_name, email}}) =>
+                            mapInstructor(user_id, `${first_name} ${last_name}`, email));
+                };
+
+                return (
+                    <div style={{width: "inherit"}}>
+                        <Grid container className="student-align">
+                            <AsyncSelect
```
dhuangg: This is a thing??? that's awesome lol.

src/components/Form/Form.js
NeilAllavarpu: I can't remove `loadInstructors` from Form.js unless Form is a functional component - it relies on `this.props.authToken`, which I need Redux to get.

Issue #204, opened 2020-01-02
Save Scheduler State in session storage


dhuangg:
After you've edited/viewed a session that's not on the current date, you'll be redirected back to the scheduler route which is the main scheduler app. The main scheduler app will by default load you back to the current date, and not the date that you looked at previously. 

For example, I'm editing a session on 1/3/2019. After I'm done editing, I expect to go back to 1/3/2019 when i go back to the scheduler tool. But as of now, I'm redirected back to the current date.

dhuangg:
I'm getting this error when I refresh
<img width="1544" alt="Screen Shot 2020-02-13 at 7 58 34 PM" src="https://user-images.githubusercontent.com/12959959/74500278-456ead00-4e9b-11ea-8052-ac4bd2226aa6.png">

PR #250, opened 2020-02-10
Scheduler state saved to session storage

NeilAllavarpu:
Addresses #204
Should save most states properly, will need to check if all states save properly.


NeilAllavarpu:
Closing for now

dhuangg:
@NeilAllavarpu were you not able to fix the bug?

PR #251, opened 2020-02-10
made some random fixes

dhuangg:
- updated search account endpoint to reflect mat's new search endpoint response
- updated weeklySessionParser() function to include the first week

### DIFF
--- a/src/reducers/searchResultsReducer.js
+++ b/src/reducers/searchResultsReducer.js
@@ -5,10 +5,11 @@ import {SEARCH_ALL} from "../actions/actionTypes";
 export default function course(state = initialState.SearchResults, {payload, type}) {
     let status = 1;
     if(payload){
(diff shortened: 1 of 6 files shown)


src/reducers/searchResultsReducer.js
```diff
@@ -5,10 +5,11 @@ import {SEARCH_ALL} from "../actions/actionTypes";
 export default function course(state = initialState.SearchResults, {payload, type}) {
     let status = 1;
     if(payload){
-        if(payload.response){
+        if( payload.response && Object.keys(payload.response).length > 0){
```
NeilAllavarpu: Is this second comparison necessary?

dhuangg:
yeah, there are cases where it breaks if there isnt a payload.response
attribute or if payload.response is an empty object. double check this for
me if you can too

Issue #175, opened 2019-11-30
Display More Search Results


dhuangg:
## Background on Search API
@tigermli was able to add an extra "page" parameter in our search endpoint, you can see this currently used in our API request configuration (SearchResults.js line 66)
```const requestConfig = { params: { query: params.query, page: 1 }, headers: {"Authorization": `Token ${props.auth.token}`,} };```

**The "page" parameter in the API request gets the first 8 results that the backend finds.** Here are the rest of the account filters + the descriptions of how they work:
- profile: possible options below
  - student
  - instructor
  - parent, 
- grade: any number between 1 and 13
  - NOTE: if this is on, no instructor, parent, receptionist, or admin should be matched. 
- sort: (options below)
  - alpha: sort alphabetically by first name
  - desc: sort by id number (descending) 

Also feel free to view [this Airtable](https://airtable.com/tblKBsoAtlwvUE8wI/viwzY9WshIXLlV0hF?blocks=hide) for more descriptions of filters for courses.

@tigermli was also kind enough to create [this documentation](https://github.com/omou-org/front-end/wiki/Search-API-Documentation) for the search API!

### Making the API Call
I already wrote the redux actions + reducer to be able to call the search endpoint. You can update the search results by calling these 2 functions:
1. `props.searchActions.fetchSearchAccountQuery(requestConfig);`
2. `props.searchActions.fetchSearchCourseQuery(requestConfig);`

`requestConfig` you see in the parameters, is the same as the one as I mentioned earlier. Here's a refresher:
```const requestConfig = { params: { query: params.query, page: 1 }, headers: {"Authorization": `Token ${props.auth.token}`,} };```

**Bottom Line**: You can activate filters by changing the requestConfig variable under "params." Just add the param/filter you want, plus the value it should be. For example, I want to filter by grade so I only see grade 2 related results, I would adjust the requestConfig like this: 
```const requestConfig = { params: { query: params.query, page: 1, grade:2 }, headers: {"Authorization": `Token ${props.auth.token}`,} };```
NOTICE: I added a "grade:2" to the "params" object

Example 2: I want to see more than the first 8 results, I would adjust the requestConfig like this:
```const requestConfig = { params: { query: params.query, page: 2 }, headers: {"Authorization": `Token ${props.auth.token}`,} };```
NOTICE: I updated the "page" param to be 2 instead of 1, to get the next 8 results. 

## What Needs To Be Done
[...]

PR #255, opened 2020-02-12
#175 display more results

dhuangg:
Starting a PR to track this more easily

### DIFF
--- a/src/components/FeatureViews/Search/SearchResults.js
+++ b/src/components/FeatureViews/Search/SearchResults.js
@@ -1,14 +1,14 @@
-import React, {useEffect, useMemo, useState} from 'react';
-import {Grid, IconButton} from "@material-ui/core";
-import {bindActionCreators} from "redux";
+import React, { useEffect, useMemo, useState } from 'react';
+import { Grid, IconButton } from "@material-ui/core";
+import { bindActionCreators } from "redux";
 import * as searchActions from "../../../actions/searchActions";
-import {connect, useDispatch, useSelector} from "react-redux";
+import { connect, useDispatch, useSelector } from "react-redux";
 import Paper from "@material-ui/core/Paper"
 import Typography from "@material-ui/core/Typography"
 import AccountsCards from "./cards/AccountsCards"
 import CoursesCards from "./cards/CoursesCards"
 import "./Search.scss";
-import {useParams} from "react-router-dom"
+import { useParams } from "react-router-dom"
 import * as apiActions from "../../../actions/apiActions";
 import * as userActions from "../../../actions/userActions";
 import * as registrationActions from "../../../actions/registrationActions";
@@ -18,7 +18,7 @@ import MoreResultsIcon from "@material-ui/icons/KeyboardArrowRight";
 import LessResultsIcon from "@material-ui/icons/KeyboardArrowLeft";
 import CourseFilters from "./CourseFilters";
 import Loading from "../../Loading";
-import {SEARCH_ACCOUNTS, SEARCH_ALL, SEARCH_COURSES} from "../../../actions/actionTypes";
+import { SEARCH_ACCOUNTS, SEARCH_ALL, SEARCH_COURSES } from "../../../actions/actionTypes";
 import Chip from "@material-ui/core/Chip";
 
 const SearchResults = (props) => {
@@ -32,7 +32,8 @@ const SearchResults = (props) => {
         }),
         [dispatch]
     );
-    const searchState = useSelector(({Search}) => Search);
+    const searchState = useSelector(({ Search }) => Search);
+    const { accounts, courses, SearchQuery, page, count } = searchState;
 
     const [start, setStart] = useState({
         account: 0,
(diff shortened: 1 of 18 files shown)


src/reducers/searchResultsReducer.js
```diff
@@ -69,36 +69,39 @@ export default function course(state = initialState.SearchResults, {payload, typ
...
+const handleAccountSearchResults = (state, payload, status) => {
+    let { response } = payload;
+    let { data } = response;
+    console.log(data)
+    // you can get page and count
     return {
         ...state,
-        accounts:data.results,
+        count: data.count,
+        page: parseInt(data.page),
```
dhuangg: page shouldn't be stored here. You should be tracking 2 pages: the current page for Account results, and the current page for Course results

src/reducers/searchResultsReducer.js
```diff
@@ -69,36 +69,39 @@ export default function course(state = initialState.SearchResults, {payload, typ
...
-
+const handleAccountSearchResults = (state, payload, status) => {
+    let { response } = payload;
+    let { data } = response;
+    console.log(data)
+    // you can get page and count
     return {
         ...state,
-        accounts:data.results,
+        count: data.count,
```
dhuangg: there eventually will be 2 counts. this count is specifically for the number of account search results. When @tigermli updates the course search endpoint, you'll also need to track course count.

src/reducers/searchResultsReducer.js
```diff
@@ -69,36 +69,39 @@ export default function course(state = initialState.SearchResults, {payload, typ
...
-const handleAccountSearchResults = (state, payload, status) =>{
-    let {response}= payload;
-    let {data} = response;
-
+const handleAccountSearchResults = (state, payload, status) => {
+    let { response } = payload;
+    let { data } = response;
+    console.log(data)
+    // you can get page and count
     return {
```
dhuangg: A better way of structuring the data should be like this:
```
...
account:{
  results: [],
  count: 0,
  page: 1,
 },
course: {
  results: [],
  count: 0,
  page: 1,
}
...
```
In this case we'll need to update this return statement and what we have in initial state to reflect this structure.

src/reducers/searchResultsReducer.js
```diff
@@ -69,36 +69,39 @@ export default function course(state = initialState.SearchResults, {payload, typ
...
 }
 
-const handleAccountSearchResults = (state, payload, status) =>{
-    let {response}= payload;
-    let {data} = response;
-
+const handleAccountSearchResults = (state, payload, status) => {
+    let { response } = payload;
+    let { data } = response;
+    console.log(data)
```
dhuangg: remove

src/components/FeatureViews/Search/SearchResults.js
```diff
@@ -43,26 +44,31 @@ const SearchResults = (props) => {
         course: 4,
     });
     const [currentPage, setCurrentPage] = useState({
```
dhuangg: Since page is something that's given to us by the backend and we're storing it in redux, we should consider getting rid of this state variable and only update the page through redux. otherwise we'd need to update it both in this component (SearchResults) and in Redux.

src/components/FeatureViews/Search/SearchResults.js
```diff
@@ -43,26 +44,31 @@ const SearchResults = (props) => {
...
             api.resetSearchParams();
         }
-    },[]);
+    }, []);
 
     const numberOfResults = () => {
         switch (searchState.primaryFilter) {
             case SEARCH_ALL: {
-                return accounts.length + courses.length;
+                return count;
```
dhuangg: Keep in mind. this is only the account count - you'll also need to add the course count when that's done developing. So be sure to add a TODO comment so you don't forget

src/components/FeatureViews/Search/SearchResults.js
```diff
@@ -140,163 +158,163 @@ const SearchResults = (props) => {
...
+                                        {
+                                            <IconButton disabled={lessResults !== 1}
+                                                className={"less"}
+                                                onClick={displayResults("less", SEARCH_ACCOUNTS)}>
+                                                <LessResultsIcon />
+                                            </IconButton>
+                                        }
+                                        {searchState.params.account.accountPage}
+                                        {
+                                            <IconButton disabled={moreResults}
```
dhuangg: instead of using a state variable for this, and you setting the state variable in the beginning, just call the comparison that you used above, but condense it. For example`<IconButton disabled={ maxPage > currentPage } ... />` and `<IconButton disabled={maxPage >= currentPage && currentPage !== 1}.../>` 

You can do this by setting only 1 state variable - `maxPage` - and deleting the other two you had - `moreResults` and `lessResults`. 

Doing this has 2 benefits:
1. Reduces the amount of code it takes to achieve this
2. Less code = more maintainable code

dhuangg:
@CalvinFronda I just fixed the redux issue. take a look at this tomorrow when you're working

dhuangg:
there's likely bugs in this PR (in particular requesting for more results), but got to get to other tickets.

PR #271, opened 2020-02-28
unenroll_CTA

dhuangg:
This adds buttons to unenroll a student from a course. We should consider refactoring RegistrationCourseEnrollments.js to pull from a course's enrollments instead of its roster - had to delete student key from roster to remove enrollment on registration course page.


src/components/FeatureViews/Accounts/TabComponents/CourseSessionStatus.js
```diff
@@ -262,6 +266,15 @@ const CourseSessionStatus = () => {
 
     };
 
+    const closeUnenrollDialog = (toUnenroll) => event => {
+        event.preventDefault();
+        setUnenrollWarningOpen(false);
+        if(toUnenroll){
+            api.deleteEnrollment(courseID, studentID, enrollment.enrollment_id);
```
NeilAllavarpu: Can you make this action take in the whole enrollment and have it extract course and student inside? It would make calling this later on simpler.

src/components/FeatureViews/Registration/RegistrationCourseEnrollments.js
```diff
@@ -37,12 +48,28 @@ const TableToolbar = (
 );
 
 const RegistrationCourseEnrollments = ({courseID}) => {
+    const dispatch = useDispatch();
+    const api = useMemo(
+        () => ({
+            ...bindActionCreators(registrationActions, dispatch),
+        }),
```
NeilAllavarpu: Since you're only using one `bindActionCreators`, you can rewrite this first parameter as:

     () => bindActionCreators(registrationActions, dispatch)

src/components/FeatureViews/Registration/RegistrationCourseEnrollments.js
```diff
@@ -37,12 +48,28 @@ const TableToolbar = (
...
     const courses = useSelector(({"Course": {NewCourseList}}) => NewCourseList);
     const parents = useSelector(({"Users": {ParentList}}) => ParentList);
     const students = useSelector(({"Users": {StudentList}}) => StudentList);
     const enrollments = useSelector(({Enrollments}) => Enrollments);
 
+    const [studentMenuAnchorEl, setStudentMenuAnchorEl] = useState(null);
+    const [unenroll, setUnenroll] = useState({
+        open: false,
+        student: null,
+        course: null,
```
NeilAllavarpu: Can you remove these fields from state and just pull it from the enrollment when needed?

src/components/FeatureViews/Registration/RegistrationCourseEnrollments.js
```diff
@@ -86,8 +113,38 @@ const RegistrationCourseEnrollments = ({courseID}) => {
...
     }
 
+    const handleClick = event => {
+        setStudentMenuAnchorEl(event.currentTarget);
+    };
+    const handleClose = event => {
+        setStudentMenuAnchorEl(null);
+    };
+
+    const handleUnenroll = (studentID, courseID, enrollment) => event => {
```
NeilAllavarpu: No need for courseID here if it is already passed as a prop. Also, extract `studentID` from `enrollment` if you can

src/components/FeatureViews/Scheduler/DisplaySessionView.js
```diff
@@ -106,6 +109,22 @@ function DisplaySessionView({ course, session, handleToggleEditing }) {
...
+       setUnenroll(true);
+    };
+
+    const closeUnenrollDialog = (toUnenroll) => event => {
+        event.preventDefault();
+        setUnenroll(false);
+        if(toUnenroll){
+            const courseID = course.course_id,
+                studentID = course.roster[0],
+                enrollmentID = enrollments[course.roster[0]][course.course_id].enrollment_id;
```
NeilAllavarpu: Add a comment to clarify that this only happens if there is one student in the roster.

src/components/FeatureViews/Accounts/TabComponents/CourseSessionStatus.js
```diff
@@ -495,7 +495,7 @@ const CourseSessionStatus = () => {
                 <DialogContent>
                     <DialogContentText>
                         {`
-                        You are currently registering ${registeringParent && registeringParent.user.name}. If you wish to continue to add sessions, you will
+                        You are currently registering ${registeringParent.user && registeringParent.user.name}. If you wish to continue to add sessions, you will
```
NeilAllavarpu: This might break if `registeringParent` is undefined/null, not sure if that is undefined/null ever. You would need all 3 checks here if so.

PR #279, opened 2020-03-05
Added PanelManager component and a testcase (TestPanel)

rgedwill:
Still cleaning up the way things are passed to the PanelManager component. I'm currently using the categories component as a skeleton, so that's why my testpanel look so similar


src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,100 @@
...
+ * Example: 
+ * { TODO}
+ * 
+ * @prop {array} operations List of operation types (i.e ["READ", "UPDATE" ) => options are: READ, UPDATE, DELETE 
+ */
+    const dispatch = useDispatch();
+    const api = useMemo(
+        () => ({
+            ...bindActionCreators(adminActions, dispatch),
+        }),
```
dhuangg: Since this is just one object, you can simplify this to:
```
const api = useMemo( () => bindActionCreators(adminActions, dispatch) );
```

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,100 @@
...
+
+    const generateGrid = () => {
+        /**
+         * Generates header with fields
+         */
+        const fieldsWithDefaults = addDefaults(props.fields);
+        const headerElements = [];
+        const records = [];
+
+        for (const [index, value] of fieldsWithDefaults.entries()) {
```
dhuangg: to use better JS, you can use the .forEach() function

example: 
```
fieldsWithDefaults.entries.forEach( ( [index, value] ) => { ... } ); 
```

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,100 @@
...
+
+            headerElements.push(
+                <Grid item xs={value["col-width"]} md={value["col-width"]} >
+                    <Typography align={value["align"]} style={{color: 'white', fontWeight: '500'}}>
+                        {value["name"]}
+                    </Typography>
+                </Grid>
+            )
+        }
+        for (const [index, value] of props.categories.entries()) {
```
dhuangg: similar story here using .forEach()

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,100 @@
...
+            <Grid container>
+                {/* header */}
+                <Grid item xs={12}>
+                    <Grid container className={'accounts-table-heading'}>
+                    {headerElements}
+                    </Grid>
+                </Grid>
+
+                {/* Directory */}
+                <Grid>
```
dhuangg: I haven't actually pulled/looked at this, but can you verify that the "Directory" is the list of items we're displaying? For example the red box highlighted here would be the directory:
Screen Shot 2020-03-05 at 4 20 16 PM
If this is it, is this the final version? If you're still working on making the update functionalities, then I'll just let you finish before I say anything, haha.

src/components/FeatureViews/AdminPortal/PanelManager.js
NeilAllavarpu: In fact, since you are trying to create a new array based on an existing one, you can use [`.map()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map):

    const headerElements = fieldsWithDefaults.map((value, index) => /* JSX */);

which can also be directly chained to the `addDefaults` call:

    const headerElements = addDefaults(props.fields).map((value, index) => /* JSX */);

src/components/FeatureViews/AdminPortal/PanelManager.js
rgedwill: Yes it is, but I need to style the records. The headings work but the records are currently just an <li>.

As for the edit function, it’s looking decent I just didn’t wanna confuse you because right now it’s too specific for the categories component, I need to rename stuff to make it more general and get it working for other field types. I’ll commit tonight

dhuangg: sounds good! rock on, man :)

NeilAllavarpu:
NOTE: change this to merge into development

PR #285, opened 2020-03-09
fixed katies bugs

CalvinFronda:
A lot in this pull:
1.  Registration Catalog filters only show on courses
2. Added new validation called "name" This helps with defining more validations to separate "short text" and other types of inputs
3. Changed student accounts tab from “Current Session” to “Current Course”
4. Added amount paid column in parents account tab


src/reducers/initialState.js
```diff
@@ -1325,7 +1325,7 @@ function initRegistrationForm() {
                     {
                         field: "Category Name",
                         name: "Category Name",
-                        type: "text",
+                        type: "shor text",
```
NeilAllavarpu: Typo?

src/reducers/initialState.js
```diff
@@ -1334,7 +1334,7 @@ function initRegistrationForm() {
                     {
                         field: "Description",
                         name: "Description",
-                        type: "text",
+                        type: "shor text",
```
NeilAllavarpu: Typo?

src/utils.js
```diff
@@ -103,3 +103,14 @@ export const truncateStrings = (string, length) => string.length > length
...
 }
+
+export const paymentToString = (string) => {
+    switch (string) {
+        case "intl_credit_card":
+            return "International Credit Card";
+        case "credit_card":
+            return "Credit Card";
+        default:
+            return string.charAt(0).toUpperCase() + string.slice(1);
```
NeilAllavarpu: Use the `capitalizeString` function.

src/components/FeatureViews/Registration/RegistrationLanding.js
```diff
@@ -189,39 +189,41 @@ const RegistrationLanding = () => {
                     </Tabs>
                 </Grid>
             </Grid>
-            <Grid
-                container
-                layout="row"
-                spacing={8}>
+            {view === 1 ? "" :
```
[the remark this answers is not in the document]
CalvinFronda: what is the second condition? why do we need an ```&&``` if we are just checking if it is on view 1 or != 1

src/components/FeatureViews/Registration/RegistrationLanding.js
```diff
@@ -187,31 +187,33 @@ const RegistrationLanding = () => {
                     </Tabs>
                 </Grid>
             </Grid>
-            <Grid
-                container
-                layout="row"
-                spacing={8}>
+            {view !== 1 ?
```
NeilAllavarpu: To make this only render when `view !== 1`, instead of using a ternary with an empty alternate value, you can simply [use the `&&` operator](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) - if the condition is false, the whole thing is replaced with `false`, and React will not render `false`, `null`, or `undefined`:

    view !== 1 && <>your JSX here</>

PR #291, opened 2020-03-11
zip utility function

rgedwill:
Identical to python's built in zip utility function


rgedwill:
No it's not a concat, it's different.
`const a = [1, 2, 3]
const b = ['a', 'b', 'c']
// concat returns:
[1, 2, 3, 'a', 'b', 'c']

//zip returns:
[ [1, 'a'], [2, 'b'], [3, 'c'] ]`

Issue #296, opened 2020-03-13
Search is broken.


dhuangg:
I really don't know what's going on anymore. 

Something is causing it to be in an infinite loop on the prod endpoint. 

## Replication instructions
1. On local.process.env - change the endpoint to production
2. Search for anything
3. Be stuck in an infinite loop

PR #300, opened 2020-03-15
Search rewrite

NeilAllavarpu:
Addresses #296


dhuangg:
Also, use AsyncCreatable instead https://react-select.com/creatable

This will allow the user to search for a random string without thinking there's no results

PR #315, opened 2020-03-21
Rewrite Admin code

NeilAllavarpu:
Not 100% done (still need to work on TuitionRules), but did most of it.

### DIFF
--- a/src/components/FeatureViews/AdminPortal/ManageCategories.js
+++ b/src/components/FeatureViews/AdminPortal/ManageCategories.js
@@ -1,272 +1,285 @@
-import React, {useEffect, useMemo, useState} from "react";
-// Material UI Imports
+import React, {useCallback, useEffect, useMemo, useState} from "react";
+import {useDispatch, useSelector} from "react-redux";
+import {bindActionCreators} from "redux";
+
+import Button from "@material-ui/core/Button";
+import EditIcon from "@material-ui/icons/Edit";
 import Grid from "@material-ui/core/Grid";
+import IconButton from "@material-ui/core/IconButton/IconButton";
+import Paper from "@material-ui/core/Paper";
+import TextField from "@material-ui/core/TextField";
+import Typography from "@material-ui/core/Typography";
 
 import "./AdminPortal.scss";
-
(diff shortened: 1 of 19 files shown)


src/reducers/adminReducer.js
```diff
@@ -46,87 +43,62 @@ export default function admin(state = initialState.Admin, { payload, type, }) {
...
-            Discounts[discountType] = Discounts[discountType].map((discount)=>{
-                if(discount.id === id){
-                    return updatedDiscount;
-                } else {
-                    return discount;
-                }
-            });
+        case "PATCH":
+            newDiscounts = Discounts[discountType].map((discount) =>
+                discount.id === id ? data : discount);
```
dhuangg: lol. i'm also dumb. i realized this was the dumb way to update an item in an array here. We should be finding the index of the discount/item to update, then only updating that specific one instead of going through the entire array. (This would be O(const) instead of O(n) I think).

src/reducers/adminReducer.js
```diff
@@ -46,87 +43,62 @@ export default function admin(state = initialState.Admin, { payload, type, }) {
...
-                if(rule.id === data.id){
-                    return updatedPriceRule;
-                } else {
-                    return rule;
-                }
-            });
-        }
+        case "PATCH":
+            newPriceRules = PriceRules.map((rule) =>
+                rule.id === data.id ? data : rule);
```
dhuangg: similar story here

src/components/FeatureViews/AdminPortal/ManageCategories.js
```diff
@@ -1,272 +1,285 @@
...
+            api.updateCategory(id, {
+                "description": editingCategory.description,
+                "id": editingCategory.id,
+                "name": editingCategory.name,
+            });
         }
         editingCategory.editing = !editingCategory.editing;
-        let updatedCategoryList = categoryList.map((category)=>{
-            if(category.id === id){
+        const updatedCategoryList = categoryList.map((category) => {
```
dhuangg: same story here

dhuangg:
NOTE: do not close this PR - still a work in progress

Issue #312, opened 2020-03-19
Removing all eslinter warnings part 1


dhuangg:
Remove all eslinter warnings for the following components

- [x] Accounts
- [x] AdminPortal
- [x] Notes
- [x] Registration

Any warning that is located in directories, reducers, or actions associated with these components is within this ticket.

PR #317, opened 2020-03-22
Accounts code cleanup

NeilAllavarpu:
Part of #312


src/components/FeatureViews/Accounts/ComponentViewer.js
```diff
@@ -1,81 +1,67 @@
...
-            component = <StudentInfo user={props.user} />;
-            break;
-        case 9:
-            component = <PayCourses user={props.user} />;
-            break;
-        default:
-            component = <InstructorSchedule />;
-    }
+const ComponentViewer = ({inView, user}) => {
+    const component = useMemo(() => {
```
dhuangg: We should change this to a JS object so that this isn't a big switch statement. This way we can keep adding new components to the components object more easily than adding a switch case

PR #318, opened 2020-03-24
cypress commands for login and finding dropdown

CalvinFronda:


cypress/integration/C_0001Sessions_paid_spec.js
```diff
@@ -0,0 +1,7 @@
+describe("Sessions Paid", () => {
+    it("Pays for session", () => {
+        cy.login({ email: "c@lvin.com", password: "password" })
```
NeilAllavarpu: Extract these to separate variables for easier reuse and changing.

cypress/support/commands.js
```diff
@@ -10,8 +10,33 @@
...
+
+/***
+ * @description: Finds a react-select dropdown when given the element and text to type
+ * @param {string} element : This is the react-select element you want to test
+ * @param {string} text  : If you want to type anything inside the react-select 
+ * @return: finds the text and selects it
+ */
+
+Cypress.Commands.add('findDropdown', (element, text) => {
+    cy.get(`${element}`).type(text ? text : "")
```
NeilAllavarpu: Can be simplified: `cy.get(element).type(text || "")`

cypress/integration/C_0001Sessions_paid_spec.js
CalvinFronda: sure

PR #319, opened 2020-03-24
set registering parent

dhuangg:


cypress/support/commands.js
```diff
@@ -15,28 +15,53 @@
...
+});
+
+/***
+ * @description: After the user as logged in, this will go to the registration tab and set the registering parent
+ * @param {string} parentName : This is the name of the parent you want to search for
+ * @return: this will set the registering parent
+ */
+
+Cypress.Commands.add("setRegisteringParent", (parentName) => {
+  cy.contains("Registration").click();
```
NeilAllavarpu: Can directly [visit](https://docs.cypress.io/api/commands/visit.html#Syntax) a URL instead of clicking on the button.

cypress/support/commands.js
```diff
@@ -15,28 +15,53 @@
...
+  cy.contains("SET PARENT").click();
+  cy.get(".select-parent-search__control").type(parentName);
+  cy.focused().type("{downarrow}{enter}", { force: true });
+  cy.get(".MuiDialogActions-root-493 > .MuiButtonBase-root-242").click();
```
NeilAllavarpu: [Use a special prop](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements) to select the button.

cypress/support/commands.js
```diff
@@ -15,28 +15,53 @@
...
+});
+
+/***
+ * @description: This will exit out of the current registering parent
+ */
+
+Cypress.Commands.add("exitRegisteringParent", () => {
+  cy.contains("Registration").click();
+  cy.get(":nth-child(3) > .MuiButtonBase-root-242").click();
```
NeilAllavarpu: See above

PR #320, opened 2020-03-24
'Add category, edit category, and add tuition rule' test

clam1987:


dhuangg:
Also, you should never branch off the branch you're currently working on (in this case your material ui branch) because now, you can only merge this PR in AFTER your material ui branch has been merged in.

PR #322, opened 2020-03-25
Add test for registering student and parent

NeilAllavarpu:


cypress/integration/B_000Student_Parent_spec.js
```diff
@@ -0,0 +1,132 @@
...
+        // Zipcode
+        cy.get(":nth-child(12) > .MuiGrid-container-141 > .MuiFormControl-root-238 > .MuiInputBase-root-273 > .MuiInputBase-input-283")
+            .click()
+            .type("94566");
+    });
+
+    it("Submits the form", () => {
+        cy.get("[style=\"margin-left: 2%;\"] > div > .MuiButtonBase-root-122")
+            .click();
+        cy.contains("successfully");
```
dhuangg: We need to do a thorough check of if this use case has been successfully submitted, so this is just the action to perform the flow, but we need to check if this flow was successful in adding a student and parent that gets reflected across the platform. When we write assertions, we should be using `cy.should()` https://codelikethis.com/lessons/javascript/cypress

We'll also need to write the actual assertion for this test by checking if we see a student and a parent appear when we search for them - to do a thorough test, we need to test to see if the student/parent ids we received can be found through Search and can we access their account pages that reflect the student/parent id's that we received from the original POST request.

PR #324, opened 2020-03-27
Panelmanager

rgedwill:


src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,306 @@
...
+            return editingRecord;
+        } else {
+            return record;
+        }
+    });
+    setRecordList(updatedRecordList);
+};
+
+const editRecordRow = (record) => {
+    console.log("edit");
```
dhuangg: Delete

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,306 @@
...
+    return (
+        <Paper square={true} className={"category-row"} >
+            <Grid container alignItems={"center"}>
+                <Grid item xs={3} md={3} >
```
dhuangg: If you have these columns fixed, then we won't be able to add additional fields (since we can only have 2 fields where 1 field takes up three columns and the other takes seven columns). The number of columns should be determined by the user.

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,306 @@
...
+        if (record.id == id) {
+            return editingRecord;
+        } else {
+            return record;
+        }
+    });
+    setRecordList(updatedRecordList);
+};
+
+const editRecordRow = (record) => {
```
dhuangg: Right now your `editRecordRow` will only ever be able to edit records that only have a "name" and a "description". This won't be the case for discounts or tuition rules. These fields should be rendered dynamically based on what key/value pairs a record has. 

Example: If we have a discount that displays "discount amount" "discount title" "discount type" etc. then this current `editRecordRow()` function won't work.

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,306 @@
...
+        for (const [index, value] of fieldsWithDefaults.entries()) {
+
+            headerElements.push(
+                <Grid item xs={value["col-width"]} md={value["col-width"]} >
+                    <Typography align={value["align"]} style={{color: 'white', fontWeight: '500'}}>
+                        {value["name"]}
+                    </Typography>
+                </Grid>
+            )
+        }
```
dhuangg: For lines 242 - 256, you should separate this to a separate `renderHeader()` function.

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,306 @@
...
+                        <Grid container spacing={8} alignItems={"center"}>
+                            {
+                                recordList.length > 0 ? recordList.map((record) => {
+                                    console.log(record);
+                                    return (<Grid item xs={12} md={12} key={record.id}>
+                                        {
+                                            record.editing ? editRecordRow(record) : viewRecordRow(record)
+                                        }
+                                    </Grid>);
+                                }): <NoListAlert list={"Course Categories"} />
```
dhuangg: `<NoListAlert/>` will display an alert stating the `list` doesn't have any item. Right now all it will display is "There are no Course Categories" even if it's a discount or a tuition rule since it's hard coded in here.  This should be a prop that's passed in.

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,306 @@
...
+                            }
+                        </Grid>
+                    </Grid>
+                </Grid>
+            </Grid>
+        )
```
dhuangg: Lines 258 - 291 should be in the last return function on its own below. having just a `generateGrid()` function in an empty React fragment seems redundant/unnecessary code.

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,306 @@
...
+    };
+
+    // const editRecord = (id) => (e) => {
+    //     e.preventDefault();
+    //     let editingRecord = records.find((record) => {return record.id === id});
+    // }
```
dhuangg: Delete comments

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,306 @@
...
+
+    return (
+        <>
+            {generateGrid()}
+        </>
+    )
+}
+
+export default PanelManager
```
dhuangg: Add React.Proptypes below this.

src/components/FeatureViews/AdminPortal/TestPanel.js
```diff
@@ -0,0 +1,121 @@
...
+    const [categoryList, setCategoryList] = useState([]);
+
+    const categories = useSelector(({Course}) => Course.CourseCategories);
+    const categoryStatus = useSelector(({RequestStatus}) => RequestStatus.category);
+
+    
+    const customUseEffects = [];
+
+    useEffect(()=>{
+        api.fetchCategories();
```
dhuangg: we should figure out if we want PanelManager to do the fetching or we want our parent component to do the fetching. Right now, we're fetching categories both in the PanelManager and this TestPanel component. I think keeping it in TestPanel would be good enough. @NeilAllavarpu what do you think?

src/components/FeatureViews/AdminPortal/customTestPanel.js
```diff
@@ -0,0 +1,48 @@
...
+        fields={fields}
+        fetchFunction={fetchFunction}
+        statusFunction={statusFunction}
+        updateFunction={updateFunction}
+        customUseEffects={customUseEffects}
+        />
+    )
+}
+
+export default CustomTestPanel
```
dhuangg: Does this actually work? There was quite a bit of hard-coded things I saw in PanelManager so I'm not sure if you try to edit one of these discounts you'd be able to edit it properly?

src/components/FeatureViews/AdminPortal/customTestPanel.js
```diff
@@ -0,0 +1,48 @@
...
+        [dispatch]
+    );
+    
+    const fields = {
+
+    };
+    const fetchFunction =[
+    api.fetchMultiCourseDiscount(),
+    api.fetchPaymentMethodDiscount(),
+    api.fetchDateRangeDiscount()
```
dhuangg: Does PanelManager handle this properly...? If I understand this correctly you will be displaying all types of discounts under 1 panel manager instead of having 3 panel managers display 3 types of discounts.

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,306 @@
...
+ **/
+    const zip = rows => rows[0].map((_, c) => rows.map(row=>row[c]));
+
+    const [categoryName, setCategoryName] = useState('');
+    const [categoryDescription, setCategoryDescription] = useState('');
+
+    //recordState
+    const [recordList, setRecordList] = useState([]);
+
+    const records = useSelector(({Course}) => Course.CourseCategories);
```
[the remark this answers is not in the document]
NeilAllavarpu: Add the selector part as a prop, e.g. `selector={({Course}) => Course.CourseCategories}` would be passed in

src/components/FeatureViews/AdminPortal/TestPanel.js
```diff
@@ -0,0 +1,121 @@
...
+    return (
+        <div>
+        {TestForm()}
+        <br />
+        <PanelManager
+            fields={fields}
+            fetchFunctions={fetchPropFunctions}
+            statusFunction={categoryStatus}
+            updateFunction={updateFunction}
+            customUseEffects={customUseEffects}
```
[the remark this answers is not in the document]
rgedwill: This is definitely what I was looking for. I was wondering if the only way to "enforce" prop structure was through docs. I was even thinking about a way to eventually make a template for the react-cli so anyone who wanted to write a new panel would be able to generate the template from the command line. This is common in vue.

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,306 @@
...
+        if(records.length !== recordList.length) {
+            let parsedRecordList = records.map((record) => ({
+                ...record,
+                editing: false,
+            }));
+            setRecordList(parsedRecordList);
+        }
+    }, [records]);
+
+    if(props.statusFunction[GET] !== 200){
```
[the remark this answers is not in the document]
rgedwill: Ok great, yeah I really need to look through the hooks.js code just to see what's available to me

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -226,60 +227,50 @@ const editRecordRow = (record) => {
...
-                                    {
-                                        record.editing ? editRecordRow(record) : viewRecordRow(record)
-                                    }
-                                </Grid>);
-                            }): <NoListAlert list={"Course Categories"} />
-                        }
-                    </Grid>
-                </Grid>
+    const headerElements = [];
+    for (const [index, value] of fieldsWithDefaults.entries()) {
```
NeilAllavarpu, line 233: Can simply use `map` on `fieldsWithDefaults.entries()`

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -226,60 +227,50 @@ const editRecordRow = (record) => {
...
+
+        headerElements.push(
+            <Grid item xs={value["col-width"]} md={value["col-width"]} >
+                <Typography align={value["align"]} style={{color: 'white', fontWeight: '500'}}>
+                    {value["name"]}
+                </Typography>
             </Grid>
```
NeilAllavarpu: JSX elements in an array should have key props

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -226,60 +283,50 @@ const editRecordRow = (record) => {
...
+        )
+    }
+
+
+    return (
+        <Grid container>
+            {/* header */}
+            <Grid item xs={12}>
+                
+                <Grid container className={'accounts-table-heading'}>
```
dhuangg, line 305: With Material UI 4, you can combine a Grid item and container like this `<Grid item xs={12} container> ... </Grid>`

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -148,15 +102,65 @@ function PanelManager(props) {
...
+                ...record,
+                editing: false,
+            }));
+            setRecordList(parsedRecordList);
+        }
+    }, [records]);
     const fieldsWithDefaults = addDefaults(props.fields);
+
+    useEffect(() => {
+        const constructDirectoryObject = (fieldInfo, recordArray) => {
```
dhuangg, line 121: Can you put some documentation of why you decided to construct the directory object this way?

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -148,15 +102,65 @@ function PanelManager(props) {
...
+        }
+
+        setDirectoryData(constructDirectoryObject(fieldsWithDefaults, records));
+    }, [records, fieldsWithDefaults])
+
+    const getRecord = (index) => Object.values(directoryData).map((field) => {
+        console.log("field")
+        console.log(field)
+        return field["values"][index]
+    })
```
dhuangg, line 143: I'm confused... this function returns a list of records for all fields at a given index? 

Are you trying to get a record from a specific index?

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -148,15 +102,65 @@ function PanelManager(props) {
...
+    const viewHeader = () => Object.values(directoryData).map((field) => {
+
+        return (
+            <Grid item xs={field["col-width"]} md={field["col-width"]} >
+                <Typography align={field.align} style={{color: 'white', fontWeight: '500'}}>
+                    {field.label}
+                </Typography>
+            </Grid>
+        )
+    })
```
dhuangg, line 154: You can clean this up like this:
```
    const viewHeader = Object.values(directoryData).map((field) => (
            <Grid item xs={field["col-width"]} md={field["col-width"]} >
                <Typography align={field.align} style={{color: 'white', fontWeight: '500'}}>
                    {field.label}
                </Typography>
            </Grid>
        )
    );
```

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -281,8 +291,18 @@ const editRecordRow = (record) => {
...
 
 
-
-
+const getRecordList = () => {
+    const rList = [];
+    for(let i = 0; i <= directoryData["ids"].length; i++ ) {
+        const record = Object.values(directoryData).map(obj => {
+            return obj.values[i];
+        });
```
dhuangg, line 299: This looks very familiar to your "getRecord()" function. Is it possible to delete the other function and just use this function? You can find the record list you're looking for then index into it? 

Still not very certain what you're trying to do for lines 297 - 299... you're saving record as a list of values from each directory data. Consider using different names or providing additional documentation.

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -281,8 +291,18 @@ const editRecordRow = (record) => {
...
+        rList.push(record);
+    }
```
dhuangg, line 301: You should be able to combine/condense your code from lines 296 to 303 to just:
```
Object.values(directoryData).forEach( (data) => rList.push( data.values ) )
```

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -141,28 +141,31 @@ function PanelManager(props) {
                         fieldInfo["ids"].push(record[key])
                     }
                 })
-                fieldInfo["editing"].push(false);
+                if(fieldInfo["editing"][index]) {
```
rgedwill, line 144: This was an attempt to only set the values to false, if they weren't already true. Not sure if this is necessary

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -205,136 +208,131 @@ function PanelManager(props) {
...
+                recordToEdit[key] = directoryData[key]["values"][index];
+            }
+  
         });
-        props.updateFunction(id, recordToUpload);
+        props.updateFunction(recordToEdit["id"], recordToEdit);
+
+        mutableDirectoryData["editing"][index] = !directoryData["editing"][index];
+        setDirectoryData(mutableDirectoryData)
+
```
rgedwill, line 257: Here, the correct value shows true. However, the end of this function causes a re-render, which sets everything to false. The editRecordRow never renders...the conditional on line 360 never returns true.

src/components/FeatureViews/AdminPortal/PanelManager.js
dhuangg: you aren't setting the value to false, you're adding a new "false" boolean to the `fieldInfo["editing"]` array. 

You can also simplify this to:
```
if(!fieldInfo["editing"][index]){
    fieldInfo["editing"][index] = false;
}
```

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,386 @@
...
+    //     setDirectoryData(updatedRecordList);
+    // };
+
+    const editRecord = (index) => (e) => {
+        e.preventDefault();
+        const recordToEdit = {};
+
+        let mutableDirectoryData = directoryData;
+        Object.keys(directoryData).map(key => {
+            if (key == "editing") {
```
dhuangg, line 243: Why do you need this case? If you don't need it then just check the other conditions:
```
key == "ids" ? recordToEdit.id = directoryData.ids.index : recordToEdit.key = directoryData.key.values.index
```

src/components/FeatureViews/AdminPortal/PanelManager.js
```diff
@@ -0,0 +1,386 @@
...
+                    if (key !== "id") {
+                        if(fieldInfo[key]["values"]){
+                            fieldInfo[key]["values"].push(record[key])
+                        } else {
+                        fieldInfo[key]["values"] = [record[key]]
+                        }
+                    } else if (key === "id") {
+                        fieldInfo["ids"].push(record[key])
+                    }
+                })
```
dhuangg, line 143: What does `recordArray` look like after this line? Does it actually reset all editing values to `false`?

src/components/FeatureViews/AdminPortal/PanelManager.js
rgedwill: They need to be false in order for the tertiary function to return the viewRecordRow don't they? Won't it complain that the value is undefined or have an index error when there's no value for fieldInfo["editing"][index] for the other rows?

dhuangg: I'm not sure what you're trying to say here (in general, avoid using "they"/"it" and specify what you're specifically referring to). 

This is how I'm understanding your code from lines 144 - 148:
1. Check if fieldInfo["editing"] is true at `index`
2. If fieldInfo["editing"] is true, then do nothing
3. If fieldInfo["editing"] is false, then add a new `false` value to the end of the fieldInfo["editing"] array

In your first comment you referenced step 1 saying
> This was an attempt to only set the values to false, if they weren't already true.

The keyword I read here was "set" - you're trying to set any non-true values to false. Based on this, I was looking for an "=" sign where you're setting the value at that index to be false. But you used `fieldInfo["editing"].push(false)` which adds a "false" value at the end of the array. This isn't the same as setting the value to false. 

Let me know if I'm missing anything here

PR #325, opened 2020-03-28
Update login test

NeilAllavarpu:


cypress/integration/A_000Login_spec.js
```diff
@@ -1,30 +1,82 @@
...
+    it("Displays error on failed submit", () => {
+        cy.server();
+        cy.route({
+            "method": "POST",
+            "response": "Invalid credentials",
+            "status": 401,
+            "url": "/auth_token/",
+        });
+        cy.get("[data-cy=signInButton]").click();
+        cy.get("[data-cy=errorMessage]");
```
dhuangg: Does this actually validate if the error message appears?

package.json
```diff
@@ -57,8 +57,9 @@
     ]
   },
   "devDependencies": {
-    "eslint": "^6.8.0",
     "cypress": "^4.2.0",
+    "eslint": "^6.8.0",
+    "eslint-plugin-cypress": "^2.10.3",
     "faker": "^4.1.0",
     "puppeteer": "^1.20.0",
```
dhuangg: we can delete faker and puppeteer

cypress/integration/A_000Login_spec.js
NeilAllavarpu: Yep. The component isn't rendered if there is no error.

PR #333, opened 2020-04-01
Unpaidstyling

gglinoga:
disregard .env.local change, was testing a bug that was fixed (ID bug)


src/components/FeatureViews/AdminPortal/UnpaidSessionCard.js
```diff
@@ -27,46 +29,54 @@ const UnpaidSessionCard = ({unpaidStudent}) => {
...
-                </Typography>
-            </CardContent>
+            <CardActionArea
+            component = {NavLink}
+            to={`/accounts/students/${studentID}/${courseID}`}
+            >
+                <CardMedia>
+                    <Grid
+                        className="unpaid-avatar-container"
+                        container>
```
dhuangg, line 44: Why do you need a <Grid container/> for this? shouldn't a div be enough?

src/components/FeatureViews/AdminPortal/UnpaidSessionCard.js
```diff
@@ -27,46 +29,54 @@ const UnpaidSessionCard = ({unpaidStudent}) => {
...
+                                "backgroundColor": adminUtils.statusColor[
+                                    unpaidStudent.sessions_left
+                                ],
+                            }}>
+                            {unpaidStudent.sessions_left}
+                        </span>
+                        <br />
+                        Amount Due: ${amtDue}
+                        <br />
+                        {course.title}
```
dhuangg, line 76: Wrap `course.title` in its own `<Typography>` tag. Give the tag a "nowrap" attribute and set it to true so that this will be an ellipses instead of wrapping on to the next line/overflowing [Typography documentation](https://material-ui.com/api/typography/). 

Then around this new `<Typography>` tag, wrap a `<Tooltip/>` that displays the entire name of the course on hover [Tooltip documentation](https://material-ui.com/api/tooltip/)

Currently still looks like this for long course names: 
Screen Shot 2020-04-04 at 2 39 01 PM

src/components/FeatureViews/AdminPortal/UnpaidSessionCard.js
gglinoga: I think there is a word missing in this comment?

gglinoga: I don't understand what this comment is asking for.

dhuangg: *Why do you need a container for this? Shouldn't a div be enough?

src/components/FeatureViews/AdminPortal/UnpaidSessionCard.js
```diff
@@ -27,41 +30,58 @@ const UnpaidSessionCard = ({unpaidStudent}) => {
 		adminUtils.calculateSessionLength(startTime, endTime)
```
NeilAllavarpu, line 30: This produces a NaN amount due for some times because you parse the course times on lines 25-26, but `calculateSessionLength` also parses the time, causing NaN to be returned here and setting `amtDue` to NaN. Remove one of the parsing calls.

PR #336, opened 2020-04-01
Moment migration

dhuangg:
migrated all of the displayed dates to use Moment instead of JS dates


dhuangg:
@NeilAllavarpu  what to do about eslint. lol

NeilAllavarpu:
looks like it's not recognizing react is being used...

PR #348, opened 2020-04-07
Dashboard

gglinoga:
Draft


dhuangg:
@gglinoga i'll do a more detailed review of your code, but at a quick glance you're definitely in the right direction for the upcoming sessions component. 

But what about the notes component? You can't just use @NeilAllavarpu 's notes component because that's a different UI. 
Screen Shot 2020-04-12 at 1 31 53 PM

I'm not sure if we covered this, but don't include the "Recently Updated Accounts" component. We're migrating to GraphQL so the implementation for that will be different with GraphQL.

src/components/FeatureViews/Dashboard/TodayCard.js
```diff
@@ -0,0 +1,44 @@
...
+import PropTypes from "prop-types";
+import {useSelector} from "react-redux";
+import Card from '@material-ui/core/Card';
+import CardContent from '@material-ui/core/CardContent';
+import Typography from "@material-ui/core/Typography";
+
+
+const TodayCard = ({session}) => {
+
+    const instructors = useSelector(({Users}) => Users.InstructorList);
```
NeilAllavarpu, line 11: Unused, remove.

src/components/FeatureViews/Dashboard/Dashboard.js
```diff
@@ -16,33 +16,36 @@ import CardMedia from '@material-ui/core/CardMedia';
...
-            <Grid xs={3}>
-                Notes
+            <Grid container justify="space-around">
+                <Grid item xs={9}>
+                    <Typography variant="h2" className="dashboard-greeting">
+                        Hello {user}!
+                    </Typography>
+                    <br/>
+                    <Typography variant='h4' className="dashboard-date">
+                        April 12, 2020
```
dhuangg, line 34: Use the [Moment library](https://www.npmjs.com/package/react-moment#formatting) to display the current date.

src/components/FeatureViews/Dashboard/DashboardNotesCard.js
```diff
@@ -0,0 +1,326 @@
...
+    const [noteTitle, setNoteTitle] = useState("");
+    const [editID, setEditID] = useState(null);
+    const [notification, setNotification] = useState(false);
+    const [submitting, setSubmitting] = useState(false);
+    const [isPost, setIsPost] = useState(false);
+    const [error, setError] = useState(false);
+    const [deleteID, setDeleteID] = useState(null);
+    const [deleteError, setDeleteError] = useState(false);
+
+    const numericDateString = (date) =>
```
dhuangg, line 62: Use the Moment component to do this

src/components/FeatureViews/Dashboard/Dashboard.js
```diff
@@ -33,62 +33,63 @@ const Dashboard = () => {
...
-                        key = {user.id}
-                        id={user.id}
-                        first_name={user.first_name}
-                    />
+                        <br/>
+                        <Paper className="today-paper" container>
+                            <Grid container style={{width: "100%", justifyContent:"space-between", padding: "5px"}}>
+                                    <Grid item xs={4}>
+                                    <Typography variant='h5' className={`dashboard-date ${classes.date}`}>
+                                        {currentDate}
```
dhuangg, line 48: As mentioned previously - use React Moment: https://www.npmjs.com/package/react-moment#formatting 

so this should look like 
```
<Moment format="dddd, MM DD">
{currentDate}
</Moment>
```

src/components/FeatureViews/Dashboard/Dashboard.js
```diff
@@ -33,62 +33,63 @@ const Dashboard = () => {
...
+                                    </Typography>
+                                    </Grid>
+                                    <Grid item xs={4}>
+                                    </Grid>
+                                    <Button 
+                                        variant="contained" 
+                                        variant="outlined"
```
dhuangg, line 55: You can't have both variants - choose one

src/components/FeatureViews/Dashboard/Dashboard.js
```diff
@@ -33,62 +33,63 @@ const Dashboard = () => {
...
+                                    <Button 
+                                        variant="contained" 
+                                        variant="outlined" 
+                                        style={{margin:"5px", float: "right"}}
+                                        component={NavLink}
+                                        to='/scheduler'
+                                        >View in Scheduler
+                                    </Button>
+                            </Grid>
+                            <Grid item xs={4}></Grid>
```
dhuangg, line 62: This can be a self-closing tag `<Grid item xs={4}/>`

dhuangg:
@gglinoga I'm getting this error when I load up your branch
Screen Shot 2020-05-10 at 10 32 58 AM

dhuangg:
hm seems to disappear when I refresh. Something to do with caching... Will keep you updated if I can figure out how to replicate it

dhuangg:
@gglinoga The css looks a lot better! There's still a couple of things that need to be tweaked though both styling and API Requests-wise
1. Styling notes
Screen Shot 2020-05-10 at 10 37 03 AM

2. Unpaid Sessions Loading: for some reason the unpaidSessionsStatus never becomes a 200 even though there was a successful request. Take a look into why unpaid sessions can't load. 

3. Upcoming Sessions Search: You're trying to call on a function in the reducer that doesn't exist. Don't just copy and paste - think about what you're trying to do here. Make sure the functions match and that you're returning a true deep copy of the state. 
Screen Shot 2020-05-10 at 10 48 45 AM

src/components/FeatureViews/AdminPortal/UnpaidSessionCard.js
```diff
@@ -71,10 +72,20 @@ const UnpaidSessionCard = ({unpaidStudent}) => {
 		unpaidStudent.sessions_left,
 		adminUtils.calculateSessionLength(startTime, endTime)
 	);
+    
+    let studentID;
+    let courseID;
+    
+    if(!student){
+    return(
+        <Loading></Loading>
```
dhuangg, line 81: change this to `<Loading small/>` to use a smaller loading image

gglinoga: View in Scheduler will view the scheduler at the previously viewed time frame.  Do you want me to force it to the today view?

dhuangg:
yes - it's confusing if you see the other views. i know it'll display the
previous scheduler view, so you'll need to 1) store the idea that the
current route is the dashboard 2) forward to the scheduler route 3) when
the scheduler component is loading, validate that if the previous route is
the dashboard; if it is then change the scheduler state to display the day
view.

src/components/FeatureViews/Scheduler/Scheduler.js
```diff
@@ -75,7 +75,9 @@ const Scheduler = () => {
     const [instructorFilter, setInstructorFilter] =
         useState(prevState.instructorFilter);
     const [timeShift, setTimeShift] = useState(prevState.timeShift || 0);
-    const [view, setView] = useState(prevState.view || "timeGridDay");
+    let timeView;
+    props.location.state ? timeView = "timeGridDay": timeView = prevState.view || "timeGridDay";
```
dhuangg, line 79: simplify line 78 and 79:
```
const timeView = props.location.state ? "timeGridDay" : prevState.view || "timeGridDay"; 
```

src/components/FeatureViews/AdminPortal/UnpaidSessionCard.js
```diff
@@ -60,32 +60,23 @@ const useStyles = makeStyles((theme) => ({
...
     const classes = useStyles();
-    const students = useSelector(({Users}) => Users.StudentList);
-    const courses = useSelector(({Course}) => Course.NewCourseList);
-
-	const student = students[unpaidStudent.student];
-	const course = courses[unpaidStudent.course];
-	const startTime = course.schedule.start_time;
-	const endTime = course.schedule.end_time;
+    
+    const studentName = `${unpaidStudent.student.user.firstName} ${unpaidStudent.student.user.lastName}`
```
dhuangg, line 64: use the fullName function in the utils file

src/components/FeatureViews/AdminPortal/UnpaidSessions.js
```diff
@@ -12,23 +15,39 @@ import "./AdminPortal.scss";
...
 const UnpaidSessions = () => {
 
-	const UnpaidList = useSelector(({Admin}) => Admin.Unpaid) || [];
-	const studentList = useMemo(() => UnpaidList.map(({student}) => student), [
-		UnpaidList,
-	]);
-	const courseList = useMemo(() => UnpaidList.map(({course}) => course), [
-		UnpaidList,
-	]);
+	const QUERIES = {
```
dhuangg, line 18: You don't need this - Neil only did this in the accounts tab because he had a lot of queries to pull from. You only need to create an UNPAID_SESSIONS_QUERY const and set that equal to what you have starting from line 19 - 39

src/components/FeatureViews/AdminPortal/UnpaidSessions.js
```diff
@@ -12,23 +15,39 @@ import "./AdminPortal.scss";
...
+		"unpaidSessions": gql`query MyQuery {
```
dhuangg, line 19: Name the query properly - this is GetUnpaidSessionsQuery, not "MyQuery"

src/components/FeatureViews/AdminPortal/UnpaidSessions.js
```diff
@@ -12,23 +15,39 @@ import "./AdminPortal.scss";
...
 
-	const studentStatus = hooks.useStudent(studentList);
-	const courseStatus = hooks.useCourse(courseList);
-	const unpaidSessionStatus = hooks.useUnpaidSessions();
+	const { data, loading, error } = useQuery(QUERIES["unpaidSessions"]);
 
-	if (hooks.isLoading(studentStatus, courseStatus, unpaidSessionStatus) && (courseStatus !== null && studentStatus !== null)) {
+	if (loading){
 		return <Loading loadingText="UNPAID SESSIONS LOADING" small/>;
 	}
```
dhuangg, line 46: Add error handling:
```
if(error) => display `error.message`
```

src/components/FeatureViews/Dashboard/Dashboard.js
```diff
@@ -37,6 +40,45 @@ const Dashboard = () => {
...
+        showFiltered: false,
+        filter: ""
+    });
+
+    const handleChange = e => {
+        e ? setCurrentFilter({filter: e.label, showFiltered: true}): setCurrentFilter({filter:"", showFiltered: false});
+    };
+
+    const QUERIES = {
+        "categories": gql`query todayCategories {
```
dhuangg, line 55: Same thing here - you don't need the QUERIES object, if you just have 1 query to call, just create a `const GET_CATEGORIES_OF_TODAY_SESSIONS = gql'query GetTodayCategories{...}'`. Again, be sure to name your variables more descriptively

src/components/FeatureViews/Dashboard/Dashboard.js
```diff
@@ -37,6 +40,45 @@ const Dashboard = () => {
...
+    }
+    
+    const { data, loading, error } = useQuery(QUERIES["categories"]);
+
+    if (loading) {
+        return (
+            <Loading/>
+        );
+    }
+
```
dhuangg, line 77: Sam here - you need to do error handling of if(error) display error.message

src/components/FeatureViews/Dashboard/Today.js
```diff
@@ -9,23 +9,55 @@ import CardContent from "@material-ui/core/CardContent";
...
 
-const Today = () => {
-    const sessionResults = useSelector(({Search}) => Search.sessions);
-    const sessionStatus = useSearchSession("", 1, "", "today", "timeAsc");
-    const sessionArray = sessionResults[1];
-    const instructorStatus = hooks.useInstructor();
-    const courseStatus = hooks.useCourse();
-    const categoryStatus = hooks.useCategory();
+const Today = (filter) => {
+    const QUERIES = {
```
dhuangg, line 17: Same thing here. don't use the QUERIES object

src/components/FeatureViews/Dashboard/Today.js
```diff
@@ -9,23 +9,55 @@ import CardContent from "@material-ui/core/CardContent";
...
+    })
 
-    if (hooks.isLoading(instructorStatus, courseStatus, sessionStatus, categoryStatus)) {
+    if (loading) {
         return (
             <Loading
                 loadingText="SESSIONS ARE LOADING"
                 small />
         );
     }
```
dhuangg, line 59: Don't forget error handling!!!

src/components/FeatureViews/Dashboard/TodayCard.js
```diff
@@ -21,27 +21,22 @@ const useStyles = makeStyles((theme) => ({
...
-    const instructors = useSelector(({Users}) => Users.InstructorList);
-    const courses = useSelector(({Course}) => Course.NewCourseList);
-    const course = session.title
-    const instructor = instructors[session.instructor].name
-    const startTime = (courses[session.course].schedule.start_time).slice(1)
-    const roster = courses[session.course].roster
-    const countStudents = roster.length
-
+    const countStudents = session.course.enrollmentSet.length
+    const instructor = `${session.course.instructor.user.firstName} ${session.course.instructor.user.lastName}`
```
dhuangg, line 25: rename this to instructorFullName and use the fullName function from utils.js

src/components/FeatureViews/AdminPortal/UnpaidSessions.js
gglinoga: Is the message "There has been an error" plus console.error sufficiently descriptive?

gglinoga: "There has been an error loading unpaid sessions"?

dhuangg: the error object has a message property so display something like `<div> There has been an error: {error.message} </div>`

src/components/FeatureViews/Dashboard/Dashboard.js
```diff
@@ -58,7 +58,7 @@ const Dashboard = () => {
                   courseCategory {
                     id
                     name
-                  }
+                  }f
```
dhuangg, line 61: cleanup

src/components/FeatureViews/AdminPortal/UnpaidSessionCard.js
```diff
@@ -61,7 +62,7 @@ const useStyles = makeStyles((theme) => ({
 const UnpaidSessionCard = ({unpaidStudent}) => {
     const classes = useStyles();
     
-    const studentName = `${unpaidStudent.student.user.firstName} ${unpaidStudent.student.user.lastName}`
+    const studentName = fullName(unpaidStudent.student.user.firstName, unpaidStudent.student.user.lastName)
```
dhuangg, line 65: Be sure to test these things... the `fullName` function takes a user object and returns `user.firstName + user.lastName` This will break the function

src/components/FeatureViews/Dashboard/TodayCard.js
```diff
@@ -22,7 +23,7 @@ const useStyles = makeStyles((theme) => ({
 const TodayCard = ({session}) => {
     const classes = useStyles();
     const countStudents = session.course.enrollmentSet.length
-    const instructor = `${session.course.instructor.user.firstName} ${session.course.instructor.user.lastName}`
+    const instructorFullName = fullName(session.course.instructor.user.firstName, session.course.instructor.user.lastName)
```
dhuangg, line 26: Same here

src/components/FeatureViews/Notes/Notes.js
```diff
@@ -64,16 +65,11 @@ const useStyles = makeStyles((theme) => ({
...
-			// transform: "scale(1)",
-			bottom: "0%",
-			right: "0%"
-		}
+        padding: "0"
 	},
 	notesNotification: {
-        float: "right",
+        // position: "absolute",
+        // float: "right",
```
dhuangg, line 72: cleanup

src/components/FeatureViews/Notes/Notes.js
```diff
@@ -292,109 +323,249 @@ const Notes = ({ownerType, ownerID}) => {
...
+                    onClose={hideWarning}
+                    open={deleteID !== null}>
+                    <DialogTitle>
+                        Confirm Delete
+                    </DialogTitle>
+                    <DialogContent>
+                        Are you sure you want to delete {
+                            notes[deleteID] && notes[deleteID].title
+                                ? `"${notes[deleteID].title}"`
+                                : "this note"
```
NeilAllavarpu, line 401: Revert this - notes can't be directly accessed by ID anymore, you have to use the `getNoteByID` function.

src/components/FeatureViews/Notes/Notes.js
```diff
@@ -292,109 +323,249 @@ const Notes = ({ownerType, ownerID}) => {
...
+                    xs={12}>
+                        <Paper elevation={2} className="note"
+                        style={{height: "150px", width: "100%"}}>
+                            <Typography
+                                align="left"
+                                className={`noteHeader ${classes.notesTitle}`}>
+                                {note.title}
+                                <Avatar
+                                    variant="square"
+                                    variant="rounded"
```
NeilAllavarpu, line 477: Only use one of these lines.

src/components/FeatureViews/Scheduler/Scheduler.js
```diff
@@ -76,7 +76,7 @@ const Scheduler = () => {
     const [instructorFilter, setInstructorFilter] =
         useState(prevState.instructorFilter);
     const [timeShift, setTimeShift] = useState(prevState.timeShift || 0);
-    const [view, setView] = useState(prevState.view || "timeGridDay");
+    const timeView = props.location.state ? "timeGridDay" : prevState.view || "timeGridDay";    const [view, setView] = useState(timeView)
```
NeilAllavarpu, line 79: This breaks currently - you should be accessing location via the `useLocation` hook from `react-router-dom`. Also, you should be checking for a specific part of the state (the `isDashboard` boolean), rather than whether or not there is a state. [Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) may be useful.

src/components/FeatureViews/Dashboard/TodayCard.js
```diff
@@ -0,0 +1,68 @@
...
+    const countStudents = session.course.enrollmentSet.length
+    const instructorObj = session.course.instructor.user;
+    const instructorFullName = fullName(instructorObj)
+    const startTime = session.course.startTime
+    const formattedStartTime = moment(startTime, "HH:mm").format("h:mm a");
+
+    return (
+        <Card className="today-card">
+            <CardActionArea
+                component = {NavLink}
```
NeilAllavarpu, line 34: Prefer `Link` instead of `NavLink` when `NavLink` isn't needed

src/components/FeatureViews/Dashboard/Today.js
```diff
@@ -0,0 +1,95 @@
...
+                <TodayCard
+                    key={session}
+                    session={session}
+                    />
+                </Grid>    
+                    )
+            )}
+        </>
+        )
+    };
```
NeilAllavarpu, line 92: `else if` not needed as this condition is mutually exclusive with the prior `if` condition.

src/components/FeatureViews/Dashboard/DashboardNotes.js
```diff
@@ -0,0 +1,50 @@
...
+    root: {
+        [theme.breakpoints.down('md')]: {
+            height: "68vh",
+            overflowY: "scroll",
+            marginTop: "72px",
+        }
+    }
+}))
+
+const DashboardNotes = (owner) => {
```
NeilAllavarpu, line 21: The (only) parameter to any functional component is `props` - did you mean to use object destructuring? If not, rename this to `props` to be clear.

src/components/FeatureViews/Dashboard/DashboardNotes.js
```diff
@@ -0,0 +1,50 @@
...
+    const ownerType = 'account'
+    const adminStatus = hooks.useAdmin();
+
+if (hooks.isLoading(adminStatus)) {
+    return (
+        <Loading
+            loadingText="NOTES ARE LOADING"
+            small />
+    );  
+}
```
NeilAllavarpu, line 34: Fetching the admin is not necessary to fetch the notes, with graphQL. Remove this.

src/components/FeatureViews/Dashboard/Dashboard.js
```diff
@@ -1,54 +1,162 @@
...
+    todayCard: {
+        [theme.breakpoints.down('md')]: {
+            fontSize: "6px"
+        }
+    }
+}))
+
+const Dashboard = () => {
+    const classes = useStyles();
+    const user = useSelector(({auth}) => auth) || [];
```
NeilAllavarpu, line 41: `auth` doesn't store all the user details, nor is it an array - it is an object formatted like so:

    {
        token
        email
        accountType
    }

You'll have to fetch other details yourself, either via the `admin` or `userInfo` fields.

src/components/FeatureViews/Dashboard/Dashboard.js
```diff
@@ -1,54 +1,162 @@
...
+    const currentDate = moment()
+    let isDisabled;
```
NeilAllavarpu, line 43: You never do anything with this variable, aside from passing it as a prop (while undefined!). Is this intended, or did you mean to initialize this to a certain value or make it a state variable?

src/components/FeatureViews/Dashboard/Today.js
```diff
@@ -0,0 +1,95 @@
...
+import './Dashboard.scss';
+import Grid from '@material-ui/core/Grid';
+import gql from "graphql-tag";
+import { useQuery } from "@apollo/react-hooks";
+
+
+const Today = (filter) => {
+    const TODAY_SESSION_QUERY= gql`
+            query todaySessionQuery($filter: String="") {
+                sessionSearch(query: $filter, time: "", sort: "timeAsc") {
```
NeilAllavarpu, line 19: Did you mean to query for only sessions from `today`?

src/components/FeatureViews/Dashboard/Dashboard.js
```diff
@@ -38,21 +39,59 @@ const useStyles = makeStyles((theme) => ({
...
+      }
+      
+      
+      
+      `;
+      
+    const user = useQuery(USER_QUERY, { 
+        variables: emailQuery, 
+    })
+
```
dhuangg, line 65: The useQuery hook, gives us an object with 3 attributes: {data, loading, error}. 

Whenever you use useQuery, always make sure loading = true before using data. 

You'll want to handle that on this line by doing this:
```
if( loading ) { //loading component } 
```

src/components/FeatureViews/Dashboard/Dashboard.js
```diff
@@ -38,21 +39,59 @@ const useStyles = makeStyles((theme) => ({
...
+    const userFirstName = user?.data?.accountSearch?.results[0]?.user?.firstName;
```
dhuangg, line 66: This isn't good - as mentioned, you should wait until data is loaded, then you don't need to do any of this. then you can use data directly.

src/components/FeatureViews/Dashboard/Dashboard.js
```diff
@@ -38,21 +39,59 @@ const useStyles = makeStyles((theme) => ({
 
 const Dashboard = () => {
     const classes = useStyles();
-    const user = useSelector(({auth}) => auth) || [];
+    const {email} = useSelector(({auth}) => auth) || [];
+    const emailQuery = { "email": email }
+    const USER_QUERY = gql`query userQuery($email: String="") {
```
dhuangg, line 44: You shouldn't need this query, just add more to the GET_ACCOUNT_TYPE query in authActions.js or just change authActions.js after line 5 to this - this includes a change to query for more information. With this extra information, you can use line 42 for more than just the email
```
const GET_EMAIL = gql`
    mutation GetDetails($token: String!) {
        verifyToken(token: $token) {
            payload
        }
    }`;

const GET_ACCOUNT_TYPE = gql`
    query GetAccountType($username: String!) {
        userInfo(userName: $username) {
            ... on AdminType {
                accountType
                user {
                    id
                    firstName
                    lastName
                    email
                }
            }
            ... on InstructorType {
                accountType
                phoneNumber
                user {
                    id
                    firstName
                    lastName
                    email
                }
            }
            ... on ParentType {
                accountType
                phoneNumber
                user {
                    id
                    firstName
                    lastName
                    email
                }
            }
        }
    }`;

export const setToken = async (token, shouldSave) => {
    try {
        const {"data": {verifyToken}} = await client.mutate({
            "context": {
                "headers": {
                    "Authorization": `JWT ${token}`,
                },
            },
            "mutation": GET_EMAIL,
            "variables": {token},
        });

        const email = verifyToken.payload.username;

        const {"data": {userInfo}} = await client.query({
            "context": {
                "headers": {
                    "Authorization": `JWT ${token}`,
                },
            },
            "query": GET_ACCOUNT_TYPE,
            "variables": {"username": email},
        });
		const {accountType, user, phoneNumber} = userInfo;

        if (shouldSave) {
            localStorage.setItem("token", token);
        }
        return {
            "payload": {
                accountType,
                email,
                token,
				user,
                phoneNumber
            },
            "type": types.SET_CREDENTIALS,
        };
    } catch (error) {
        // invalid token, do nothing
        console.error(error);
        return {
            "type": null,
        };
    }
};

export const logout = () => {
    localStorage.removeItem("token");
    client.clearStore();
    return {
        "type": types.LOGOUT,
    };
};
```

src/components/FeatureViews/Dashboard/Today.js
dhuangg: yes - @gglinoga this should only before today. otherwise you're going to get all the sessions that aren't today. this is the dashboard.

PR #349, opened 2020-04-07
initial commit

clam1987:


clam1987:
Hey Neil, I'm done with displaying the search results based on the filter. I did encounter a bug however. When I initially search under a filter i.e. all, I get 4 results as intended. When I switch the filter to accounts/courses and do a search for the exact keywords, I still get 4 results. When I refresh the page the filter then gets set and it displays 8 results. This also happens when you start at accounts/course and change the filter to all. I think it's the useMemo but I'm not 100% sure.

PR #355, opened 2020-04-27
Cypress testing aa

clam1987:
Can't find the issue so I'll just comment here. This PR is for AF_00Add_Profile_Notes. If I find the issue, I'll come back and edit this to tag that issue.

### DIFF
--- a/.env.local
+++ b/.env.local
@@ -1,3 +1,3 @@
 #NOTE: only variables starting with "REACT_APP_" will be read
-#REACT_APP_DOMAIN = http://localhost:8000
-REACT_APP_DOMAIN = http://api.omoulearning.net
+REACT_APP_DOMAIN = http://localhost:8000
+#REACT_APP_DOMAIN = http://api.omoulearning.net
--- a/cypress/fixtures/profile.json
+++ b/cypress/fixtures/profile.json
@@ -1,5 +1,10 @@
 {
-  "id": 8739,
-  "name": "Jane",
-  "email": "jane@example.com"
-}
+  "admin": {
+  "email": "maggie@summit.com",
+  "password": "password"
+  },
+  "receptionist": {
+  "email": "test",
+  "password": "test2"
+  }
+ }
--- a/cypress/integration/AF_00Add_Profile_Notes_Spec.js
+++ b/cypress/integration/AF_00Add_Profile_Notes_Spec.js
@@ -12,37 +12,33 @@ describe("AF_00Add_Profile_Notes", () => {
     });
     it("Will create a note with a subject and body", () => {
         cy.get(".addNote").click();
-        cy.contains("Save").should("be.disabled")
-        cy.get(".textfield").type("Out of office");
-        cy.get(".note-body").type("Daniel has coronavirus and will not be in class for the new few days");
+        cy.contains("Save").should("be.disabled");
+        cy.get(".textfield").type("Out of office for today");
+        cy.get(".note-body").type("Daniel has some things to take care of and will not be coming in today");
         cy.contains("Save").click();
-        cy.get(".noteHeader").contains("Out of office");
-        cy.get(".body").contains("Daniel has coronavirus and will not be in class for the new few days")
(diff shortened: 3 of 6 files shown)


cypress/integration/Add_profile_spec.js
```diff
@@ -0,0 +1,83 @@
...
+    it("Will display the number of important notes on the notes badge", () => {
+        cy.get("[data-cy=note-tab]").last().should('have.text', '1 Notes');
+    })
+    it("Will delete all notes", () => {
+        cy.get("[data-cy=delete-btn]").last().click();
+        cy.get("[data-cy=delete-confirm]").click();
+        cy.get("[data-cy=delete-btn]").first().click()
+        cy.get("[data-cy=delete-confirm]").click();
+    });
+    it("Will go into the registrations after login, click on AP Computer Science, and click on notes", () => {
```
CalvinFronda, line 44: If this is in the same test, why do we need to login again?

clam1987: when using cy.visit(), it reloads the page so if we don't login the test does not run.

cypress/integration/Add_profile_spec.js
```diff
@@ -0,0 +1,83 @@
...
+    });
+    it("Will create a note with a subject and body", () => {
+        cy.get("[data-cy=add-note]").click();
+        cy.contains("Save").should("be.disabled");
+        cy.get("[data-cy=textfield]").type("Problem with Stephen Colbert");
+        cy.get("[data-cy=note-body]").type("Stephen is a good kid but he keeps making too many jokes and distracting the class, his work is declining in quality");
+        cy.contains("Save").click();
+        cy.get("[data-cy=note-header]").contains("Problem with Stephen Colbert");
+        cy.get("[data-cy=body]").contains("Stephen is a good kid but he keeps making too many jokes and distracting the class, his work is declining in quality");
+    });
```
dhuangg, line 60: As mentioned previously on our call - this isn't necessary since you tested this earlier. delete lines 52 - 60

cypress/integration/Add_profile_spec.js
```diff
@@ -0,0 +1,83 @@
...
+        cy.contains("Save").should("be.disabled");
+        cy.get("[data-cy=textfield]").type("Stephen's Improvement");
+        cy.get("[data-cy=note-body]").type("He has taken things more seriously in the past few weeks and his grades have improved immensely!");
+        cy.get("[data-cy=dialog-notification]").click();
+        cy.contains("Save").click();
+        cy.get("[data-cy=note-header]").contains("Stephen's Improvement");
+        cy.get("[data-cy=body]").contains("He has taken things more seriously in the past few weeks and his grades have improved immensely!");
+        cy.get("[data-cy=edit-btn]").last().click();
+        cy.get("[data-cy=dialog-notification]").should('have.css', 'color', 'rgb(255, 0, 0)');
+        cy.contains("Cancel").click();
```
dhuangg, line 72: This is the same - this has already been tested earlier. delete 61 - 73

cypress/integration/Add_profile_spec.js
```diff
@@ -0,0 +1,83 @@
...
+    });
+    it("Will check the notes tab if the notes icon has a yellow svg", () => {
+        cy.get("[data-cy=note-number-of-icons]").last().should('exist');
+    })
```
dhuangg, line 76: This is the same - this has already been tested earlier. delete.

cypress/integration/Add_profile_spec.js
```diff
@@ -0,0 +1,83 @@
...
+    it("Will delete all notes", () => {
+        cy.get("[data-cy=delete-btn]").last().click();
+        cy.get("[data-cy=delete-confirm]").click();
+        cy.get("[data-cy=delete-btn]").first().click();
+        cy.get("[data-cy=delete-confirm]").click();
+    });
```
dhuangg, line 82: This is the same. delete

PR #366, opened 2020-04-28
Finished AC_00Courses_Save_Note_spec

clam1987:
added 'data-cy' attributes to 'Notes.js' on lines 376,484, and 489. Changed 'profile.json' to include login information.

### DIFF
--- a/.env.local
+++ b/.env.local
@@ -1,3 +1,3 @@
 #NOTE: only variables starting with "REACT_APP_" will be read
-#REACT_APP_DOMAIN = http://localhost:8000
-REACT_APP_DOMAIN = http://api.omoulearning.net
+REACT_APP_DOMAIN = http://localhost:8000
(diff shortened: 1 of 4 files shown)


dhuangg:
@clam1987 remember our discussion when we were working on this last time? Since the notes UI is the same for accounts, courses, and enrollments, we should just have 1 set of tests that test the UI, and 3 separate tests that focus on just checking to see if the notes component/tab work for accounts/courses/enrollments. 

Right now, this PR looks almost identical to #355 when really the only difference is just this from this PR. We should have 3 separate "Will go to the ____, to check if notes component load" tests (i.e. one for account notes, course notes, enrollment notes)
```
    it("Will go into the registrations, click on AP Computer Science, and click on notes", () => {
        cy.get('a[href*="/registration"]').click();
        cy.contains('AP Computer Science').click();
        cy.contains("Notes").click();
    });
```

PR #367, opened 2020-04-28
M 01 02 add category edit category delete category

clam1987:
Branch is tagged wrong but test for category M is completed.


cypress/integration/M_00Add_Edit_Category_spec.js
```diff
@@ -0,0 +1,34 @@
+describe("M_01_Add_Category_spec", () => {
+	it("Login", () => {
+		cy.fixture("profile").then((user) => {
+			cy.login({email: user.admin.email, password: user.admin.password});
+		});
+	});
+	it("Will go to the admins page and click on manage course", () => {
+		cy.get("a[href*='/admin']").click();
```
dhuangg, line 8: use `cy.visit(url)` to go to a link instead: https://docs.cypress.io/api/commands/visit.html#Syntax

clam1987: Ok so I changed line 8 to 'cy.visit(admin portal's url)' but the way I structured the test should account for error handling. Or in this case, should I be using a different assertion and not 'should'? The way I structured the test, it should type in the description first and check if the button is still disabled so you can't save it. If that case passes, it moves on to type in the title and then proceeds to save it. I've noticed that when I edit the category, I can leave everything blank and hit update and before refresh it'll be blank:
https://gyazo.com/d416c852605103a0c8b9bd0fc086426c
When I hit refresh however, it corrects itself.

dhuangg: your screenshot isn't showing. I'm not sure what you're asking about - you're combining a lot of issues into 1 paragraph so it's a bit difficult to follow. 

Most of what you said sounds good, but if i'm understanding you correctly you're having an issue when you're trying to edit a category?

cypress/integration/M_00Add_Edit_Category_spec.js
```diff
@@ -1,11 +1,9 @@
 describe("M_01_Add_Category_spec", () => {
-	it("Login", () => {
+	it("Will login and go to the admins page and click on manage course", () => {
 		cy.fixture("profile").then((user) => {
 			cy.login({email: user.admin.email, password: user.admin.password});
 		});
-	});
-	it("Will go to the admins page and click on manage course", () => {
-		cy.get("a[href*='/admin']").click();
+		cy.visit("http://localhost:3000/adminportal")
```
dhuangg, line 6: You don't need the "localhost:3000" it's already the root url

cypress/integration/M_00Add_Edit_Category_spec.js
clam1987: Actually, I think I got it correct the first time. I was just typing out loud lol. The bug I was talking about was related to editing a category. Once the category is created you can go back in and edit the name and details. You can, for instance, just delete the name + the description and hit save and it will save as empty fields. However, when you refresh the page, it defaults back to the un-editted category. I don't see this is a HUGE issue but it is a bug non the less.

dhuangg:
we're using react admin to add categories so we won't need this test anymore. closing this PR

Issue #363, opened 2020-04-27
Tooltip for Important Note


dhuangg:
As per Selina's comment
> When adding a new note, would recommend including a tooltip on hover over this icon so users know what the purpose of it is and that it is indeed clickable

## Development
Add a [Material Tooltip](https://material-ui.com/components/tooltips/#tooltip) over the icon with label "This is an Important Note!"

PR #378, opened 2020-05-06
Added Tooltip to 'UserProfile.js' and 'RegistrationCourse.js'

clam1987:
Gave the tooltip title of "This is an Important Note". #363


CalvinFronda:
The ticket is asking, "When adding a new note". What you have here is adding the tooltip on the tabs from the userProfile. We need to edit the notes.js file, around line 467

CalvinFronda:
Wilson.. please...
Add the tool tip to the icon. Not the save button. 

```When adding a new note, would recommend including a tooltip on hover over this icon so users know what the purpose of it is and that it is indeed clickable```

PR #383, opened 2020-05-13
Account edit cta

rgedwill:
**NOTE:** I think it would be good to write the tests for this before it's merged, just need to clarify how exactly I'd go about doing that... do I need to make a user story, or just create a test spec for Accounts?

Either way, the ticket itself is done.


rgedwill:
Not sure why those other commits are tacked onto this push...Maybe because I branched from my last cypress branch. For some reason my master branch is broken

Issue #359, opened 2020-04-27
Broken Back button - Edit Session -> Display Session


dhuangg:
As per Selina's comments
> Clicking the “Back” button the page took me back to the main Scheduler page instead of the course details I was just viewing

### Flow to correct to
We want to make sure when we're on the "Edit Session" page (as displayed below) and we press "Back", we go back to the second screenshot
1) Press "Back"
<img width="1466" alt="Screen Shot 2020-04-26 at 8 15 02 PM" src="https://user-images.githubusercontent.com/12959959/80330803-0fcf3400-87fb-11ea-89fb-19569ca7730b.png">
2) The page that should be displayed when "Back" is clicked
<img width="1466" alt="Screen Shot 2020-04-26 at 8 18 26 PM" src="https://user-images.githubusercontent.com/12959959/80330817-1bbaf600-87fb-11ea-8018-45abc5fecad9.png">

## Background
As of now, we're just flipping a variable "edit" to see if we need to display the "EditSessionView" or the "DisplaySessionView". The URL/route does not change. 
```
    const handleEditToggle = (cancel) => (event) => {
        event.preventDefault();
        if (!cancel && edit) {
            // if we're applying to edit session then toggle to edit view
            handleToggleEditing(editSelection);
        } else {
            setEdit(!edit);
        }
    };
```

## Development
When "Edit Session" is clicked, we should redirect it to a new route. Once this is redirected to a new route, the back button will work as intended.

Issue #372, opened 2020-04-29
Clean up Account Cards


dhuangg:
As per Selina's Comments
> For accounts, the name cards need to be consistent in size. If we want to make more room for longer names, then we need to increase the size for all account cards
> Add more padding between the initials icon circle and the account details

<img width="620" alt="Screen Shot 2020-04-28 at 7 24 03 PM" src="https://user-images.githubusercontent.com/12959959/80555682-d7a82c80-8985-11ea-98b4-c69493e37758.png">

Issue #358, opened 2020-04-27
Add Cancel Edit Session Button


dhuangg:
As per Selina's comments
> Page needs a Cancel button in case user does not want to apply any edits to the session/course
<img width="1466" alt="Screen Shot 2020-04-26 at 8 15 02 PM" src="https://user-images.githubusercontent.com/12959959/80330707-b8c95f00-87fa-11ea-9462-120692bd3ca2.png">

PR #384, opened 2020-05-14
selinas ticketz

jsflca2012:
😎


CalvinFronda:
1. This does not fix #359, still takes me back to scheduler not the session im looking at. 
2. #358 👍 LGTM
3. #372 Is there a reason we have truncated names instead of the full name? 
Screen Shot 2020-05-15 at 4 40 44 PM

dhuangg:
#372 is tricky... I'm not sure how to handle it. Try reading this to see if it offers any good suggestions: 
https://uxdesign.cc/its-all-in-the-cards-ux-ui-card-design-44cf9e31d988
https://www.justinmind.com/blog/cards-ui-design/

jsflca2012:
> 1. This does not fix #359, still takes me back to scheduler not the session im looking at.
> 2. #358 👍 LGTM
> 3. #372 Is there a reason we have truncated names instead of the full name?
>    Screen Shot 2020-05-15 at 4 40 44 PM
> 
> @dhuangg @jsflca2012

Weird... this fixed #359 for me, I'll double check later.

#372 is truncated since it prevents the card size from being overflowed by the length of the person's name.

dhuangg:
@jsflca2012 let's not truncate the names - it somewhat defeats the point of searching for it. I think for all cards, let's set the height so that by default it'll always give at least 2 line-height worth of spacing. Anything longer than 2 lines we'll truncate + add a tooltip to reveal the full name/text

jsflca2012:
READY FOR REVIEW @CalvinFronda 😎

CalvinFronda:
#359 Looks good


#372 Im not sure this is what you were expecting. 
This is in AccountsCards line 66, you have a ```height: 23vh``` which would be good if we were viewing this page on mobile.

With that in mind maybe we can use media queries to change the hight on md screens and below. Check out this material ui breakpoints guide. https://material-ui.com/customization/breakpoints/ You're going to need to use ```makeStyles```. If you need help lmk.

one more thing, AccountsCards line 82. Replace ```xs={8}``` to ```xs={12}``` in the Grid item. this will give more space for the email to spread out.

Screen Shot 2020-05-30 at 9 42 41 PM

CalvinFronda:
Looks good, last thing we need to fix is the course cards. Give each card an 
```item xs={12} align="left"``` to give each description an entire line to display the content. 

Screen Shot 2020-06-03 at 1 07 38 PM

After this we should close this and be done with it 😄

**Comment by jsflca2012** [comment]
pushed up a fix, not sure if it works since my search is a bit broken right now.

**Comment by CalvinFronda** [comment]
You added the ```xs={12}``` but not the ```align="left"```. We need to align left because the website is naturally ```align=center```.  After that we should be able to merge

**Comment by jsflca2012** [comment]
@NeilAllavarpu @dhuangg can you check this out, it's reporting a syntax error for ";" but everything looks good to me, I might just be making a dumb mistake but just want to confirm.

**Comment by dhuangg** [comment]
@jsflca2012 you didn't fully migrate the back button to be a functional component. `label() { ... }` is [the syntax for writing a _method_ in a javascript class](https://medium.com/predict/javascript-functions-vs-methods-and-other-helpful-tips-e58a621b1d27), so it won't work in a functional component. You need to convert the following class methods to functions (arrow functions):
1. `label()`
2. `handleClick()`
3. `renderIcon()`

Since this is a functional component, we no longer need the `render()` method on line 80. Just return whatever you want to render. Example:
this ✅ 
```
const BackButton = () => {
   return <button>Back Button</button>
};
```
not ❌ 
```
const BackButton = () => {
   render(){ <button>Back Button</button>}
};
```

**Comment by jsflca2012** [comment]
Crap, I didn't notice that Calvin had migrated the component to be a functional component, so when I merged my changes it it just completely exploded. Everything should be fixed now, let me know if the button is buggy.

File: src/components/BackButton.js [hunk 1]
```diff
@@ -18,24 +18,23 @@ const BackButton = (props) => {
...
 
-	handleClick() {
-		if (this.props.warn) {
-			this.setState({
-				alert: true,
-			});
+
+	const handleClick = () => {
+		if (props.warn) {
+			alert=true;
```
CalvinFronda, line 37: If we are trying to set state, it should be ```setAlert(true)```.

jsflca2012:
Yes, I believe at the time we were designing this button, we had confirmaction and denyaction functions because we were planning to have custom actions for them, but as far as I'm aware, no development has been made on them. There are supposed to be crucial functions though, such as saveform, which saves the form data before exiting to the previous page. I'll just leave them on here for now as future-proofing.

src/components/BackButton.js
```diff
@@ -34,7 +34,7 @@ const BackButton = (props) => {
 
 	const handleClick = () => {
 		if (props.warn) {
-			alert=true;
+			setAlert=true;
```
CalvinFronda, line 37: Incorrect. 
When we are using functional components we don't set state by ```=```. **Look at the differences**
 
Functional component 
```
const Backbutton = ( ) => {
      const [alert, setAlert] = React.useState(false)
    
         setAlert(true)
    }
```
Notice how I don't need to use ```this.setState```, we just use the 2nd item in that array as a function. 

Class Component  
```
class BackButton extends React.Component{
    constructor(props){
        super(props);
        this.state = {
            alert : false
        }
    }
    this.setState({alert: true})
}

```

Notice how this is so much more code for the same thing.

When working with React states you will always either of the two.

❗ **Set the state in the state in the BackButton component with the function component way** ❗ 

source: https://reactjs.org/docs/state-and-lifecycle.html

PR #390, opened 2020-05-31
Accounts graphql

NeilAllavarpu:
Notes and some of accounts i guess

### DIFF
--- a/src/components/FeatureViews/Notes/Notes.js
+++ b/src/components/FeatureViews/Notes/Notes.js
@@ -21,9 +21,9 @@ import LoadingError from "../Accounts/TabComponents/LoadingCourseError";
 import NotificationIcon from "@material-ui/icons/NotificationImportant";
 import Paper from "@material-ui/core/Paper";
 import TextField from "@material-ui/core/TextField";
+import Tooltip from "@material-ui/core/Tooltip";
 import Typography from "@material-ui/core/Typography";
 import {makeStyles} from "@material-ui/core/styles";
-import Tooltip from "@material-ui/core/Tooltip";
 
 import "./Notes.scss";
 import "../Accounts/TabComponents/TabComponents.scss";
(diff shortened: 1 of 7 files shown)


src/components/FeatureViews/Notes/Notes.js
```diff
@@ -15,503 +17,462 @@ import EditIcon from "@material-ui/icons/EditOutlined";
...
-		month: "numeric",
-	});
+const numericDateString = (date) => new Date(date).toLocaleTimeString("en-US", {
+    "day": "numeric",
+    "hour": "2-digit",
+    "minute": "2-digit",
+    "month": "numeric",
+});
+
+const QUERIES = {
```
dhuangg, line 60: Can't these queries be pulling from a fragment instead of repeated like this?

src/components/FeatureViews/Notes/Notes.js
```diff
@@ -15,503 +17,462 @@ import EditIcon from "@material-ui/icons/EditOutlined";
...
+                break;
+            }
+            default: {
+                const note = {
+                    "body": noteBody,
+                    "complete": false,
+                    "important": notification,
+                    "title": noteTitle,
+                    "user": ownerID,
+                };
```
dhuangg, line 253: Shouldn't we create a default note object and then add an additional key+pair to the note object depending on which case it is? (i.e. add "user" if it's an account note, add "course" if it's a course note, etc.)

src/components/FeatureViews/Notes/Notes.js
```diff
@@ -15,503 +17,462 @@ import EditIcon from "@material-ui/icons/EditOutlined";
...
+            </Dialog>
+            <Grid item md={3}>
+                <div className="addNote" onClick={openNewNote}
+                    style={{"cursor": "pointer"}}>
+                    <Typography className="center">
+                        <AddIcon /><br />Add Note
+                    </Typography>
+                </div>
+            </Grid>
+            {notes && Object.values(notes).map((note) => (
```
dhuangg, line 428: is the `notes &&` necessary now that we're waiting for the `loading` var to be true?

src/components/FeatureViews/Notes/Notes.js
NeilAllavarpu: I'll change this in a later PR when implementing mutations.

PR #394, opened 2020-06-05
Forgot password

NeilAllavarpu:
Views for the forgot password / reset password


src/components/Authentication/ResetPassword.js
```diff
@@ -0,0 +1,98 @@
...
+        "color": theme.colors.darkGray,
+        "margin-bottom": "30px",
+        "margin-top": "20px",
+    },
+}));
+
+const ResetPassword = () => {
+    const [password, setPassword] = useState("");
+    const [submitted, setSubmitted] = useState(false);
+    const email = "nelson@summit.com";
```
dhuangg, line 30: Can we change this to a state var? For testing, we can set "nelson@summit.com" as a default

NeilAllavarpu: This was a dummy email - when the full reset functionality is implemented, this will be automatically set from the backend/link.

PR #397, opened 2020-06-08
Smorg

clam1987:
Smorg PR

### DIFF
--- a/src/components/FeatureViews/AdminPortal/AdminDashboard.js
+++ b/src/components/FeatureViews/AdminPortal/AdminDashboard.js
@@ -1,0 +1,576 @@
+import React, { useState, useEffect, useMemo } from "react";
+import {
+  Chart,
+  BarSeries,
+  PieSeries,
+  Title,
+  ArgumentAxis,
+  ValueAxis,
+  Legend,
+  Tooltip,
+} from "@devexpress/dx-react-chart-material-ui";
+import { EventTracker } from "@devexpress/dx-react-chart"
+// import { Chart, SeriesTemplate, CommonSeriesSettings, Title } from 'devextreme-react/chart';
+import { Animation, LineSeries } from "@devexpress/dx-react-chart";
+import { makeStyles, withStyles } from "@material-ui/core/styles";
+import Typography from "@material-ui/core/Typography";
+import { ThemeProvider } from "@material-ui/styles";
+import { createMuiTheme } from "@material-ui/core";
+import AppBar from "@material-ui/core/AppBar";
+import Toolbar from "@material-ui/core/Toolbar";
+import Card from "@material-ui/core/Card";
+import Tabs from "@material-ui/core/Tabs";
+import Tab from "@material-ui/core/Tab";
+import Box from "@material-ui/core/Box";
+import Paper from "@material-ui/core/Paper";
+import Grid from "@material-ui/core/Grid";
+import ChromeTabs from "./ChromeTabs";
+import TabPanel from "./TabPanel";
+import ProfileCard from "../Accounts/ProfileCard";
+import UserAvatar from "../Accounts/UserAvatar"
+import "./AdminPortal.scss";
+import {useSelector} from "react-redux";
+
+const baseTheme = createMuiTheme();
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    flexGrow: 1,
+    // backgroundColor: theme.palette.background.paper,
+  },
+  paper: {
+    padding: theme.spacing(2),
+    textAlign: "center",
+    color: theme.palette.text.secondary,
+  },
+  title: {
+    color: "#404143",
+  },
+  tabName: {
+    float: "left",
+    marginLeft: "24px",
+    marginTop: "10px",
+    marginBottom: "10px",
+    fontWeight: "bold",
+    lineHeight: "28px",
+    fontFamily: "Roboto",
+    fontStyle: "normal",
+    color: "#666666",
+  },
+  chromeTabStart: {
+    alignSelf: "flex-end",
+    border: "1px solid #EEEEEE",
+    borderBottom: "0px",
+    borderLeft: "0px",
+    borderTopLeftRadius: 8,
+    borderTopRightRadius: 8,
+    backgroundColor: "#ffffff",
+  },
+  chromeTabEnd: {
+    alignSelf: "flex-end",
+    border: "1px solid #EEEEEE",
+    borderBottom: "0px",
+    borderRight: "0px",
+    borderTopLeftRadius: 8,
+    borderTopRightRadius: 8,
+    backgroundColor: "#ffffff",
+  },
+  chromeTab: {
+    alignSelf: "flex-end",
+    border: "1px solid #EEEEEE",
+    borderBottom: "0px",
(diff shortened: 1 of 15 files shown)


src/components/FeatureViews/AdminPortal/AdminDashboard.js
```diff
@@ -171,6 +173,47 @@ const AdminDashboard = (props) => {
...
+        }
+        lastPaidSessionDatetime
+      }
+    }
+    `,
+  }
+
+// console.log(UNPAID_SESSIONS)
+
+ const unpaidSessionsData = useQuery(QUERIES["unpaidsessions"])
```
dhuangg, line 200: separate this out to `const {data, loading, error}` remember in my demo how I handled the loading logic here. `loading` is a boolean, so if it's true then display a loading component.

src/components/FeatureViews/AdminPortal/AdminDashboard.js
```diff
@@ -171,6 +173,47 @@ const AdminDashboard = (props) => {
...
+//  console.log(unpaidSessionsData)
+ 
+ 
+ const calculateUnpaidSessions = data => {
+   if(unpaidSessionsData.data) {
```
dhuangg, line 205: You shouldn't need this. there's the loading variable i just mentioned. so before this, you should have a line of code like `if(loading) <Loading/>` or something similiar

src/components/FeatureViews/AdminPortal/AdminDashboard.js
```diff
@@ -197,69 +198,78 @@ const AdminDashboard = (props) => {
...
- const unpaidSessionsData = useQuery(QUERIES["unpaidsessions"])
-//  console.log(unpaidSessionsData)
+
+
+
+ const {data, loading, error} = useQuery(QUERIES["unpaidsessions"])
+//  console.log(data)
+ 
  
 const getTime = (time) => {
```
dhuangg: What type of time is this? This is a bit confusing. UTC time? Pacific time? minutes? use a more descriptive variable name

src/components/FeatureViews/AdminPortal/AdminDashboard.js
```diff
@@ -197,69 +198,78 @@ const AdminDashboard = (props) => {
...
+  const minutes = parseInt(time.slice(-2), 10) / 60;
+  const hours = parseInt(time.substring(0,2), 10);
   return hours + minutes;
 }; 
 
- const calculateUnpaidSessions = () => {
-   if(unpaidSessionsData.data) {
-     const TotalAmountDue = unpaidSessionsData.data.unpaidSessions.map(time=> {
-     const { endTime, startTime, hourlyTuition } = time.course
+ const unpaidSessionsObj = () => {
```
dhuangg: So this is an array of unpaid sessions (not an "Obj")? I'm a bit confused since you're returning parentObj (which is an array, since data.unpaidSessions.map() returns an array) on line 259, so what are you doing here? 

It looks like you're returning an array of unpaid sessions based on `data.unpaidSessions`. Rename your variables so that this is more clear. Also, if you're just returning `parentObj` you can just return your `data.unpaidSessions.map()`, so it would just be `const unpaidSessionsList = data.unpaidSession.map(...);`

src/components/FeatureViews/AdminPortal/AdminDashboard.js
```diff
@@ -197,69 +198,78 @@ const AdminDashboard = (props) => {
...
+  return parentObj;
+ };
+
+// const grabUnpaidSessionsObj = () => {
+//   const fullName = data.unpaidSessions.map(parents => 
+//   }))
+//   return fullName
+// }
+
+const unpaidSessionsCard = () => {
```
dhuangg: uhhh what. so this is just the `unpaidSessionsObj` array?

src/components/FeatureViews/AdminPortal/AdminDashboard.js
dhuangg: A review of array's .map() function would be good 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

src/components/FeatureViews/AdminPortal/AdminDashboard.js
```diff
@@ -171,6 +173,47 @@ const AdminDashboard = (props) => {
     { label: "Access Control" },
   ];
   
+  const QUERIES = {
```
dhuangg, line 179: I forgot to note this last time - Don't copy what Neil did in Accounts for this. Since you're just making 1 query (GET_UNPAID_SESSIONS), just make a `const GET_UNPAID_SESSIONS` query and set that equal to `gql'query...`.

src/components/FeatureViews/AdminPortal/AdminDashboard.js
```diff
@@ -171,6 +173,47 @@ const AdminDashboard = (props) => {
...
+    "unpaidsessions": gql`query MyQuery {
```
dhuangg, line 180: Be sure to rename `MyQuery` to something more descriptive (e.g. "GetUnpaidSessions")

src/components/FeatureViews/AdminPortal/AdminDashboard.js
```diff
@@ -229,47 +242,55 @@ const AdminDashboard = (props) => {
...
+       const amountDue = dollarsPerSession * Math.ceil(missedPaymentSessions)
+       const firstName= user.paymentList[0].parent.user.firstName;
+       const lastName= user.paymentList[0].parent.user.lastName;
+       const id= user.paymentList[0].parent.user.id;
+       return {name: `${firstName} ${lastName}`, initial: `${firstName.charAt(0)}${lastName.charAt(0)}`, id: id, due: amountDue}
+    });
+    return parentObj || [];
+   };
+
+  const filterAndAddDuplicates = arr => {
```
dhuangg, line 268: Can you put some documentation above this line to describe what you're doing here? You have your comments below, you can put that in the documentation. This is an example of documentation/how you should format it:
```
/**
 * Pads a number to the desired length, filling with leading zeros
 * @param {Number} integer Number to pad
 * @param {String} length Minimum number of digits
 * @returns {String} Padded integer
 */
```

src/components/FeatureViews/AdminPortal/AdminDashboardCards.js
```diff
@@ -0,0 +1,94 @@
...
+    <Paper elevation={0}>
+      <Typography className={classes.snapName}>{props.snapName}</Typography>
+      <br />
+      <br />
+      <Typography className={classes.number}>{props.number}</Typography>
+    </Paper>
+  );
+};
+
+export const OutstandingPaymentCard = (op) => {
```
dhuangg, line 61: don't abbreviate!!! write out the whole name to clarify what you mean. This is also the props for outstanding payment, so I'm not sure why you're calling all of the props as "op"

dhuangg: you should also just destructure from the prop like `({initials, name, amtDue})` instead of `(op)`

CalvinFronda:
@clam1987 is this branch still active? If so can you fix the merge conflicts?

PR #400, opened 2020-06-15
Authentication

NeilAllavarpu:
Migration to graphQL authentication


NeilAllavarpu:
Migrated search - there will probably be more small things that use the REST endpoints that will be broken, but enough of the features should be converted to be usable for further development.

PR #402, opened 2020-06-20
Cart button A08

jsflca2012:


src/components/FeatureViews/Registration/RegistrationLanding.js
```diff
@@ -97,98 +107,118 @@ const RegistrationLanding = () => {
...
-            <RegistrationActions/>
+            <Grid container>
+                <Grid item md={1}>
+                    <BackButton />
+                </Grid>
+                <Grid item md={10}/>
+                <Grid item md={1}>
+                    <Button component={NavLinkNoDup}
+				    to="/registration/cart">
+                        <Badge badgeContent={renderCartItems()} showZero color="primary" >
```
CalvinFronda, line 214: This could be simplified by:
```
<Badge badgeContent={numToCheckout === null ? 0 : numToCheckout} .../>
```
and we could remove the ```renderCartItems``` function.

jsflca2012:
> I get a bug that crashes my browser when I click on the cart with nothing inside it. Do we have anything in the design to redirect users to a "no items in cart" page?

No, originally we handled this by not making the checkout page clickable until we had courses selected. I can either fix this by making cart unclickable until items are in the cart, or handle it via the checkout page. What do you think?

CalvinFronda:
Hmm make it unclickable for now but lets talk with the design team for some insight.

src/components/FeatureViews/Registration/RegistrationLanding.js
dhuangg: @jsflca2012 this still hasn't been addressed

PR #405, opened 2020-06-21
reorganized misc components to "OmouComponents" dir, created new back…

dhuangg:
…ground paper component, lessons learned - don't messed with the paper component's root styling


src/components/OmouComponents/AddSessions.js
```diff
@@ -1,7 +1,7 @@
 import {useDispatch, useSelector} from "react-redux";
 import React, {useMemo, useState} from "react";
 import {bindActionCreators} from "redux";
-import * as registrationActions from "../actions/registrationActions";
+import * as registrationActions from "../../actions/registrationActions";
```
NeilAllavarpu, line 4: this is why we have absolute imports

src/components/OmouComponents/BackgroundPaper.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import Paper from "@material-ui/core/Paper";
+
+export default function BackgroundPaper({children, ...rest}) {
+	return (<Paper style={{"padding": "3%"}} {...rest}>
```
NeilAllavarpu, line 5: Does this need `className="paper"` or not?

src/components/OmouComponents/AddSessions.js
dhuangg: lol my IDE breaks with absolute imports. It can only automatically refactor files if its a relative import, so i had to manually fix all of the absolute imports :(

src/components/OmouComponents/BackgroundPaper.js
dhuangg: eh..... I want to delete that class. I think it's too confusing between the `<Paper>` component and this `<BackgroundPaper>` component.

PR #423, opened 2020-07-26
Notification settings + some access control

dhuangg:


src/components/Authentication/LoginPage.js
```diff
@@ -90,14 +90,17 @@ const LoginPage = () => {
...
-        login({
+        const loginResponse = await login({
             "variables": {
                 password,
                 "username": email,
             },
         });
+        if (loginResponse?.data?.tokenAuth) {
+            history.push("/")
+        }
```
NeilAllavarpu, line 103: Are we sure that redux auth state is fully loaded when the redirect occurs?

src/components/FeatureViews/Accounts/ComponentViewer.js
```diff
@@ -25,56 +25,67 @@ const ComponentViewer = ({inView, user}) => {
...
+				id: 10,
 			}
 		].filter((tab) => (tab.access_permissions.includes(AuthUser.accountType))),
 		[user]
 	);
 
 	return (
 		<Grid className="profile-component-container">
-			{componentsArray[inView].component}
+			{componentsArray.find(({id}) => id == inView).component}
```
NeilAllavarpu, line 88: Is there a reason for this change?

src/components/Authentication/LoginPage.js
dhuangg: it looks like it - i'm not getting any errors when i was testing this

src/components/FeatureViews/Accounts/ComponentViewer.js
dhuangg: We're filtering down the componentsArray based on which has access to which tabs/component. When we filter it down, the ordering of the components don't match the tab ordering from UserProfile.js (e.g. if Notes has index 5 in UserProfile.js, it may not have index 5 in ComponentViewer when we filter down the array of components there). This way we can specify exactly which tab we're looking for

PR #425, opened 2020-07-26
Parent invite

NeilAllavarpu:


src/components/FeatureViews/Accounts/TabComponents/StudentInfo.js
```diff
@@ -1,55 +1,92 @@
...
-		).isRequired,
-	}).isRequired,
+const StudentInfo = () => {
+    const {accountID} = useParams();
+    const {data, loading, error} = useQuery(GET_STUDENTS, {
+        "variables": {"id": accountID},
+    });
+
+    const classes = useStyles();
+
```
dhuangg, line 54: Do additional access control here to verify the current user is either the student's parent or is a receptionist or admin user. If the user isn't a parent or a receptionist, be sure they can't access the `studentInvite` link (aka, set `studentInvite` to false when it's not one of these 2)

Issue #429, opened 2020-07-27
Pre-populate category field in AdminPanel's Tuition Rules edit panel.


rgedwill:
On edit of a tuition rule in the AdminPanel, the category field does not pre-populate. 

The only successful way to get it to prepopulate thus far is to change the source of the edit component (TuitionEdit in AdminPanelViews.js) from "category" to "category.id". While this successfully prepopulates the field, it will break when trying to submit the change. 

The reason it breaks is because instead of passing in the category ID as an int in the GQL query, it passes the entire category object. When the source attribute is just "category" it correctly passes only the ID as an int.

rgedwill:
Still improvements to be made, but for now the issue is resolved by working around the dataprovider specifically for the tuitionRules panel

PR #431, opened 2020-07-31
Hotfixed issue #429

rgedwill:


src/components/FeatureViews/AdminPortal/dataProvider.js
```diff
@@ -442,7 +441,6 @@ export default {
                 "data": Object.values(Object.values(response.data)[0])[0],
             };
         } catch (error) {
-            console.error(error);
```
CalvinFronda, line 443: We should probably keep this error catch.

dhuangg:
@rgedwill can you follow-up on this?

CalvinFronda:
I believe this is old and has been fixed in another branch - Closing

PR #435, opened 2020-08-04
Fixed Styling & added tests

CalvinFronda:
This PR refers to Bug hunting event A005 - No results found styling needs fixing

I've removed the scss file and converted all styles with materialUI makeStyles.

Currently the login spec inside cypress is broken. I made a temporary login tests within NoResultsPage_spec.


CalvinFronda:
#455

PR #527, opened 2020-08-04
Account bug fixes

CalvinFronda:
Account List and Grid view are smaller #455 
Registration - Fixed spacing between setting a parent and shopping cart icon #482 

Fixed format to match exactly as the figma shows. #484 

edit: wrong link

### DIFF
--- a/src/components/OmouComponents/BackButton.js
+++ b/src/components/OmouComponents/BackButton.js
@@ -1,12 +1,11 @@
 import React from "react";
-import {withRouter} from "react-router-dom";
+import { withRouter } from "react-router-dom";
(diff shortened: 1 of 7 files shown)


src/components/OmouComponents/BackButton.js
```diff
@@ -80,7 +79,7 @@ const BackButton = (props) => {
 		<Hidden mdDown>
 			<Button
 				className="control course button"
-				onClick={()=>{handleClick()}}
+				onClick={() => { handleClick() }}
```
dhuangg, line 82: change to `onClick={handleClick}`

PR #533, opened 2020-08-07
Hamburger menu color

rgedwill:


rgedwill:
Issue #440 (A004)

CalvinFronda:
Its still black

Screen Shot 2020-08-06 at 10 45 14 PM

If its just a color change, I'd just change it with inline styles like this. ```style={{ color: "white" }}```

dhuangg:
sorry, i didnt read this carefully enough. @Ryan if this is outside of the
Mui Theme Provider it may be why the color isn't being applied. if that's
the case then just push up @Calvin's temp fix

dhuangg:
@rgedwill this is still black for me 
Screen Shot 2020-08-10 at 5 39 01 PM

Issue #506, opened 2020-08-04
A079 - Tuition says zero when registering class as parent


dhuangg:
A079
tuition says zero when registering class as parent
Bug found by: Katie
Teams link: katiebelleho: 

when registering as parent, im not able to change the sessions

 

…

Issue #503, opened 2020-08-04
A076 - Cannot finalize payment if total amount is zero (after credits/adjustments)


dhuangg:
A076
cannot finalize payment if total amount is zero (after credits/adjustments)
Bug found by: Katie
Teams link: katiebelleho: ah. then if the total comes out to zero, how do i finalize my registrat…

Issue #492, opened 2020-08-04
A065 - Duplicate enrollments are happening


dhuangg:
A065
Duplicate enrollments are happening
Bug found by: Wilson
Teams link:

Issue #480, opened 2020-08-04
A052 - Cart button not updating after quick registering


dhuangg:
A052
Cart button not updating after quick registering
Bug found by: Daniel
Teams link: https://teams.microsoft.com/l/message/19:77d621b48fd54d03ba7155385d33cb54@thread.skype/1596333115409?tenantId=4757b031-c3cc-4b2a-ac2c-a2ef4731ed22&groupId=7f2d402f-bd5b-4f3d-875f-fb78e2889c1a&parentMessageId=1596333115409&teamName=Omou&channelName=Defects&createdTime=1596333115409

dhuangg:
need to change the architecture/data flow to use redux instead so that the useSelector() hooks can trigger a re-render with the updated values. 

1. [connect react-final-form to redux](https://codesandbox.io/s/4xq2qpzw79?file=/src/finalFormDuck.js)
2. update the `submit` functions in FormFormats
3. update the setParentDialog handler functions to dispatch redux actions instead
4. update RegistrationActions and CourseList to subscribe to useSelector to get latest registrationCartState

PR #534, opened 2020-08-07
registration related bugs

### LINKED ISSUE #523: A096 - Able to register without selecting student
A096
Able to register without selecting student
Bug found by: Gina
Teams link: https://teams.microsoft.com/l/message/19:77d621b48fd54d03ba7155385d33cb54@thread.skype/1596321034037?tenantId=4757b031-c3cc-4b2a-ac2c-a2ef4731ed22&groupId=7f2d402f-bd5b-4f3d-875f-fb78e2889c1a&parentMessageId=1596321034037&teamName=Omou&channelName=Defects&createdTime=1596321034037

dhuangg:
this PR should have closed
#523 
#520 
#506 
#504 
#503 
#494
#492 
#480 
#479 
#469 

also fixed an unrelated unpaid sessions card bug that's breaking development


src/components/FeatureViews/Registration/RegistrationCart/PaymentBoard.js
```diff
@@ -213,21 +296,27 @@ export default function PaymentBoard() {
 	const handlePayment = async () => {
 		const paymentMethod = paymentMethodState.find(({checked}) => checked)?.value;
 		const {data: {enrollments}} = enrollmentResponse;
+		const isSameEnrollment = ({enrollment, course, student}) =>
+			(`${enrollment.student.user.id}-${enrollment.course.id}` === `${student}-${course}`);
+		console.log({enrollments}, "previous enrollments");
```
CalvinFronda, line 301: remove log

src/components/FeatureViews/Registration/RegistrationCart/PaymentBoard.js
```diff
@@ -213,21 +296,27 @@ export default function PaymentBoard() {
...
+
 		const enrollmentsToCreate = classRegistrations
 			.filter(({course, student}) =>
-				(!enrollments.find(enrollment =>
-					(course === enrollment.course.id && student === enrollment.student.user.id))))
+				(!enrollments.some((enrollment) => {
+					console.log(isSameEnrollment({enrollment, course, student}));
```
CalvinFronda, line 306: remove log

src/components/FeatureViews/Registration/RegistrationCart/RegistrationCartContainer.js
```diff
@@ -87,21 +86,36 @@ export default function RegistrationCartContainer() {
...
 	});
 
 	useEffect(() => {
-		if (!coursesResponse.loading && Object.values(registrationCartState).length > 0) {
-			const courseData = coursesResponse.data.courses;
+		dispatch({
+			type: types.INIT_COURSE_REGISTRATION,
+			payload: {},
+		})
+	}, []);
```
CalvinFronda, line 112: Not sure, but do we add dependency's into these useEffect hooks? 

like ```[types.INIT_COURSE_REGISTRATION]```, correct me if I'm wrong

dhuangg: yeah, we do. i just wanted to confirm that this would run on page load. nothing should change if i put it in though. good catch

src/components/Form/Fields.js
```diff
@@ -62,7 +62,7 @@ export const DataSelect = ({ request, optionsMap, name, ...props }) => {
     });
 
     const renderOption = useCallback(
-        ({ label }) => <span data-cy={`${name}-${label}`}>{label}</span>,
+        ({label, value}) => <span data-cy={`${name}-${value}`}>{label}</span>,
```
NeilAllavarpu, line 65: Is there a reason for doing this by option value instead of display label?

src/reducers/registrationReducer.js
```diff
@@ -61,19 +68,23 @@ export default function registration(
     case actions.POST_DISCOUNT_PAYMENT_METHOD_SUCCESS:
       return successSubmit(state);
     case actions.SET_PARENT:
-      newState.CurrentParent = payload;
-      return {...newState};
+      setParentRegistrationCart(payload);
```
NeilAllavarpu, line 71: Reducers are supposed to be [never perform side effects](https://redux.js.org/basics/reducers#handling-actions). Try moving this to the action instead.

src/components/FeatureViews/Registration/RegistrationActions.js
```diff
@@ -37,44 +34,56 @@ query GetRegisteringParent($userId: ID!) {
...
-	const numberOfRegistrationsInCart = updatedRegistrationCartNum || Object.values(registrationState).reduce((accumulator, currentStudent) => {
-		return accumulator + currentStudent.length;
-	}, 0);
+	const {
+		submitStatus = {},
+		...registrationCartState
+	} = registrationState;
+	const numberOfRegistrationsInCart = Object.values(registrationCartState).length > 0 ?
+		Object.values(registrationState).reduce((accumulator, currentStudent) => (accumulator + currentStudent?.length),
+			0) : 0;
```
NeilAllavarpu, line 86: Ternary not needed, this will simplify to 0 if registrationState is empty anyways

src/components/FeatureViews/Registration/RegistrationCart/PaymentBoard.js
```diff
@@ -213,21 +296,23 @@ export default function PaymentBoard() {
 	const handlePayment = async () => {
 		const paymentMethod = paymentMethodState.find(({checked}) => checked)?.value;
 		const {data: {enrollments}} = enrollmentResponse;
+		const isSameEnrollment = ({enrollment, course, student}) =>
+			(`${enrollment.student.user.id}-${enrollment.course.id}` === `${student}-${course}`);
```
NeilAllavarpu, line 300: Wouldn't it be clearer to compare `student === enrollment.student.user.id` and `course === enrollment.course.id`?

src/components/FeatureViews/Registration/RegistrationCart/PaymentBoard.js
```diff
@@ -213,21 +296,23 @@ export default function PaymentBoard() {
 	const handlePayment = async () => {
 		const paymentMethod = paymentMethodState.find(({checked}) => checked)?.value;
 		const {data: {enrollments}} = enrollmentResponse;
+		const isSameEnrollment = ({enrollment, course, student}) =>
```
NeilAllavarpu, line 299: Instead of making the parameter a single object, you could simply have 3 different parameters.

src/components/FeatureViews/Registration/RegistrationCart/PaymentBoard.js
```diff
@@ -260,9 +345,9 @@ export default function PaymentBoard() {
 		history.push(`/registration/receipt/${payment.data.createPayment.payment.id}`)
 	};
 
-	if (error) {
-		console.error(error.message);
-		return <div>There has been an error! : {error.message} </div>
+	if (error || enrollmentResponse.error) {
+		console.error(error.message, enrollmentResponse.error.message);
+		return <div>There has been an error! : {error.message} | {enrollmentResponse.error.message}</div>
```
NeilAllavarpu, line 350: Did you mean to write `{error.message || enrollmentResponse.error.message}` instead? As it is, if one errors but not the other, there will be an extra `|`.

src/components/FeatureViews/Registration/RegistrationCart/PaymentBoard.js
```diff
@@ -337,8 +423,9 @@ export default function PaymentBoard() {
 				<Button
 					variant="contained"
 					color="primary"
-					disabled={priceQuote.total <= 0 || priceQuote.total === "-"}
+					disabled={priceQuote.total === "-" || priceQuote < 0}
```
NeilAllavarpu, line 426: Why is priceQuote able to be `"-"` as well as a number?

src/components/FeatureViews/Registration/RegistrationCart/RegistrationCartContainer.js
```diff
@@ -87,21 +86,36 @@ export default function RegistrationCartContainer() {
 	const studentIds = (Object.keys(registrationCartState).length > 0 && Object.keys(registrationCartState)) ||
 		currentParent.studentList;
 	// create list of courses to fetch
-	const courseIds = [].concat.apply([], Object.values(registrationCartState))
-		.map(({course}) => course.id);
-	const {data, loading} = useQuery(GET_STUDENT_INFOS, {variables: {userIds: studentIds}});
-	const coursesResponse = useQuery(GET_COURSES_TO_REGISTER, {variables: {courseIds: courseIds}});
+	const courseIds = Object.values(registrationCartState).filter(reg => reg).length > 0 &&
+		[].concat.apply([], Object.values(registrationCartState))
```
NeilAllavarpu, line 90: Is this supposed to flatten the array? Try `arr.flat()` instead.

src/components/FeatureViews/Registration/RegistrationCart/RegistrationCartContainer.js
```diff
@@ -87,21 +86,36 @@ export default function RegistrationCartContainer() {
 	const studentIds = (Object.keys(registrationCartState).length > 0 && Object.keys(registrationCartState)) ||
 		currentParent.studentList;
 	// create list of courses to fetch
-	const courseIds = [].concat.apply([], Object.values(registrationCartState))
-		.map(({course}) => course.id);
-	const {data, loading} = useQuery(GET_STUDENT_INFOS, {variables: {userIds: studentIds}});
-	const coursesResponse = useQuery(GET_COURSES_TO_REGISTER, {variables: {courseIds: courseIds}});
+	const courseIds = Object.values(registrationCartState).filter(reg => reg).length > 0 &&
```
NeilAllavarpu, line 89: This condition is not really necessary, as the following value defaults to an empty array if there are no non-empty registrations. However, you will have to check for an array of length 0 instead of `false` when seeing if any IDs are present.

src/components/FeatureViews/Registration/RegistrationCart/RegistrationCartContainer.js
```diff
@@ -154,7 +172,9 @@ export default function RegistrationCartContainer() {
 		}
 	}
 
-	if (loading || coursesResponse.loading) return <Loading small/>;
+	if (loading || !data) return <Loading small/>;
+	if (error) return <div>There's been an error:
+		{error.message}</div>
```
NeilAllavarpu, line 177: Combine this into one line for clarity

src/components/FeatureViews/Registration/RegistrationCart/StudentRegistrationsEntry.js
```diff
@@ -14,27 +14,37 @@ import TextField from "@material-ui/core/TextField";
...
 	const handleSessionCheckChange = (_) => {
 		updateSession(numSessions, !checked, studentId, course.id);
-		removeRegistration(studentId, course.id);
+		dispatch({
+			type: types.DELETE_COURSE_REGISTRATION,
+			payload: {
+				studentId,
+				courseId: course.id,
+			}
+		});
```
NeilAllavarpu, line 43: We should be migrating this to graphQL at some point.

src/components/FeatureViews/Registration/SelectParentDialog.js
```diff
@@ -65,24 +58,14 @@ const SelectParentDialog = ({onClose, open, updateCartNum}) => {
...
-				setRegisteringParent(
-					JSON.parse(sessionStorage.getItem("registrations"))?.currentParent
-				)
-			);
-		}
-	}, [currentParent, dispatch]);
-
 	useEffect(() => {
 		if (!getSavedParentCartResult?.loading && getSavedParentCartResult?.called) {
 			if (getSavedParentCartResult.data) {
```
NeilAllavarpu, line 63: It seems you might want the `onCompleted` option of `useLazyQuery` here

src/reducers/registrationReducer.js
```diff
@@ -223,63 +234,9 @@ const failedSubmit = (state) => ({
...
-      "class",
-      enrollmentObject
-  );
-  prevState.submitStatus = "success";
-
-  return {...prevState};
+const addClassRegistration = ({courseId, studentId}) => {
+  const newRegistrationState = submitRegistration(studentId, courseId);
+  return {...newRegistrationState};
 };
```
NeilAllavarpu, line 240: See above. Perhaps the action can send the updated state as payload

src/components/FeatureViews/Registration/CourseList.js
```diff
@@ -59,45 +63,81 @@ const CourseList = ({ filteredCourses, updatedParent }) => {
...
+            type: types.ADD_CLASS_REGISTRATION,
+            payload: {
+                studentId: quickStudent,
+                courseId: quickCourseID
+            }
+        });
         setOpen(false);
     }
 
+    const disableQuickRegister = ({course, enrolledCourseIds, registrations, studentList}) => {
```
NeilAllavarpu, line 123: Rename to `shouldDisableRegister` to clarify that this returns a boolean instead of performing an action.

src/index.js
```diff
@@ -69,19 +67,17 @@ if (token) {
...
     </Provider>,
     document.getElementById("root"),
 );
 
-// expose store when run in Cypress
-if (window.Cypress) {
-    window.store = store;
-}
+// expose store
+window.store = store;
```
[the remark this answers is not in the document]
dhuangg: i did this intentionally - I'm not convinced that we have anything sensitive going into our redux store that wouldn't already be going through network traffic. But more importantly, I need to dispatch redux actions on submit of our form component. I haven't been able to find a good way of doing this other than this. Let me know if you can figure out a better way of dispatching actions from FormFormats

src/components/Form/Fields.js
dhuangg: this is for testing purposes - it's difficult to select items by label but values, you can always assume it's by an id that starts from 1, so it's easier to select that.

src/components/FeatureViews/Registration/RegistrationCart/PaymentBoard.js
dhuangg: i was desperate. i was trying to figure out why i couldn't filter out existing enrollments to prevent double enrollments. :( i'll change it back though lol

src/components/FeatureViews/Registration/RegistrationCart/PaymentBoard.js
dhuangg: this is a trick - it allows you to give a function all of your parameters such that order doesn't matter. so if i call: `isSameEnrollment({ newEnrollment, course, student})` I wouldn't get a bug if I did `isSameEnrollment({ course, newEnrollment, student})`

Only started doing this because i spent an hour hunting down a bug where i switched up the parameter order -_-

src/components/FeatureViews/Registration/RegistrationCart/PaymentBoard.js
dhuangg: i was only thinking about the situation if there are 2 errors. I think of some better way of handling this.

src/components/FeatureViews/Registration/RegistrationCart/PaymentBoard.js
dhuangg: visually, the default is a "-", and whenever i'm getting the price quote from the backend it becomes a number

src/components/FeatureViews/Registration/RegistrationCart/RegistrationCartContainer.js
dhuangg: there's an extra submitStatus key that gets generated into the registrationCartState when you pass it through redux. I couldn't find where to get rid of it, so i'm just filtering it out since it's value is null

src/components/FeatureViews/Registration/RegistrationCart/StudentRegistrationsEntry.js
dhuangg: This isn't to delete a course registration in the database. it deletes it from the parent's registration cart which is why i need to use redux to do this

PR #549, opened 2020-08-22
scheduler popovers

CalvinFronda:
#476

### DIFF
--- a/cypress/integration/Scheduler/Scheduler_filter_spec.js
+++ b/cypress/integration/Scheduler/Scheduler_filter_spec.js
@@ -1,0 +1,20 @@
+describe("Tests Scheduler Filter if react-select throws 'options.reduce is not a function' error", () => {
+    before(() => {
+        cy.visit("scheduler")
+        cy.login("danielhuang@blah.com", "password")
+    })
+    it("Should go to scheduler page", () => {
+        cy.visit("scheduler")
+        cy.get("[data-cy=scheduler-header-text]").contains("Scheduler")
+    })
+    it("Checks if you can enter your own filter", () => {
+        cy.get("[data-cy=schduler-select-filter-box]").click()
+        cy.get("#filter-instructor-select").type("TESTINGSTESTING").type('{enter}')
+        cy.get("#filter-instructor-select").click(0, 20)
+        cy.get("#filter-course-select").type("TEST2TEST2TEST2").type('{enter')
+        cy.get("#filter-instructor-select").click(0, 20)
+        cy.get("#filter-instructor-select").should('have.value', '')
+        cy.get("#filter-instructor-select").click(0, 20)
+        cy.get("#filter-course-select").should('have.value', '')
+    })
+})
--- a/src/components/FeatureViews/Scheduler/SessionFilters.js
+++ b/src/components/FeatureViews/Scheduler/SessionFilters.js
@@ -1,10 +1,10 @@
-import React, {useState} from "react";
+import React, { useState } from "react";
 import IconButton from "@material-ui/core/IconButton";
 import FilterIcon from "@material-ui/icons/FilterList";
 import Menu from "@material-ui/core/Menu";
-import {withStyles} from "@material-ui/core/styles";
+import { withStyles } from "@material-ui/core/styles";
 import ReactSelect from "react-select";
-import {MenuItem, Tooltip} from "@material-ui/core";
+import { MenuItem, Tooltip } from "@material-ui/core";
 
 const styles = {
 	root: {
@@ -13,13 +13,13 @@ const styles = {
 };
 
 function SessionFilters({
-							onInstructorSelect,
-							InstructorValue,
-							InstructorOptions,
-							CourseValue,
-							onCourseSelect,
-							CourseOptions,
-						}) {
+	onInstructorSelect,
+	InstructorValue,
+	InstructorOptions,
+	CourseValue,
+	onCourseSelect,
+	CourseOptions,
+}) {
 	let [anchorEl, setAnchorEl] = useState(null);
 	let [open, setOpen] = useState(false);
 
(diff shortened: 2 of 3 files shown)


src/components/FeatureViews/Scheduler/SessionFilters.js
```diff
@@ -61,20 +63,22 @@ function SessionFilters({
 					// styles={customStyles}
 					placeholder={"Filter Instructor..."}
 					value={InstructorValue}
-					options={InstructorOptions}
+					options={InstructorOptions || []}
```
NeilAllavarpu, line 66: Set the default value of InstructorOptions to be the empty array `[]` instead of doing that here.

src/components/FeatureViews/Scheduler/SessionFilters.js
```diff
@@ -61,20 +63,22 @@ function SessionFilters({
...
 					isMulti
+					id="filter-instructor-select"
 				/>
 				<ReactSelect
 					className={"session-filter-select"}
 					// styles={customStyles}
 					placeholder={"Filter Course..."}
 					value={CourseValue}
-					options={CourseOptions}
+					options={CourseOptions || []}
```
NeilAllavarpu, line 77: See above

cypress/integration/Scheduler/Scheduler_filter_spec.js
```diff
@@ -0,0 +1,20 @@
...
+    })
+    it("Checks if you can enter your own filter", () => {
+        cy.get("[data-cy=schduler-select-filter-box]").click()
+        cy.get("#filter-instructor-select").type("TESTINGSTESTING").type('{enter}')
+        cy.get("#filter-instructor-select").click(0, 20)
+        cy.get("#filter-course-select").type("TEST2TEST2TEST2").type('{enter')
+        cy.get("#filter-instructor-select").click(0, 20)
+        cy.get("#filter-instructor-select").should('have.value', '')
+        cy.get("#filter-instructor-select").click(0, 20)
+        cy.get("#filter-course-select").should('have.value', '')
```
NeilAllavarpu, line 18: Try to use the `data-cy` attribute when selecting for specific elements

CalvinFronda: Valid, but we are using react-select for these elements. I've tried: 

```
<ReactSelect
      data-cy="filter-instructor-select"
      className={"session-filter-select"}
      placeholder={"Filter Instructor..."}
      value={InstructorValue}
      options={InstructorOptions}
      onChange={onInstructorSelect}
      clearable
      isMulti
     />
```

This doesn't work so I've added the id prop and that works. Let me know if you have a solution to test React-admin.

src/components/FeatureViews/Scheduler/SessionFilters.js
CalvinFronda: I did this: 
```
    const instructorOptions = useMemo(() => {
        if (!instructors) return []
        Object.entries(instructors).map(
            ([instructorID, instructor]) => ({
                "label": instructor.name,
                "value": instructorID,
            })

        )
    }, [instructors]);
```

src/components/FeatureViews/Scheduler/SessionFilters.js
CalvinFronda: Same as above

cypress/integration/Scheduler/Scheduler_filter_spec.js
clam1987: So what you should be able to do is traverse the Virtual DOM if you add a data-cy attribute to the parent. You would use  something like .children() to traverse it. Source [https://docs.cypress.io/api/commands/children.html#Syntax](https://docs.cypress.io/api/commands/children.html#Syntax). There is also .next() [https://docs.cypress.io/api/commands/next.html#Syntax](https://docs.cypress.io/api/commands/next.html#Syntax)

cypress/integration/Scheduler/Scheduler_filter_spec.js
```diff
@@ -0,0 +1,20 @@
+describe("Tests Scheduler Filter if react-select throws 'options.reduce is not a function' error", () => {
+    before(() => {
+        cy.visit("scheduler")
+        cy.login("danielhuang@blah.com", "password")
+    })
+    it("Should go to scheduler page", () => {
+        cy.visit("scheduler")
+        cy.get("[data-cy=scheduler-header-text]").contains("Scheduler")
```
clam1987, line 8: Change contains to an assertion. Use .should() instead. Something like this:
```
cy.get("[data-cy=scheduler-header-text]").should("have.value", "Scheduler")
```

cypress/integration/Scheduler/Scheduler_filter_spec.js
clam1987: Also should lines 12-18 have more assertions so we could actually test things? I.E
```
cy.get("#filter-instructor-select").type("TESTINGSTESTING").type('{enter}').should("have.value", '{enter}')
```

also, fix line 14.
```
 cy.get("#filter-course-select").type("TEST2TEST2TEST2").type('{enter} <---')
```

cypress/integration/Scheduler/Scheduler_filter_spec.js
```diff
@@ -0,0 +1,20 @@
+describe("Tests Scheduler Filter if react-select throws 'options.reduce is not a function' error", () => {
+    before(() => {
+        cy.visit("scheduler")
+        cy.login("danielhuang@blah.com", "password")
+    })
```
NeilAllavarpu, line 5: Change this to use the `visitAuthenticated` command

src/components/FeatureViews/Scheduler/SessionFilters.js
NeilAllavarpu: Sounds good, is it included in this PR? Don't forget to return the mapped array as well.

PR #550, opened 2020-08-23
Fixed #486

### LINKED ISSUE #486: A059 - When you enter an invalid email and click sign in, then login with a valid email it will login without needing to press signin
A059
when you enter an invalid email and click sign in, then login with a valid email it will login without needing to press signin
Bug found by: Jeffrey
Teams link: https://teams.microsoft.com/l/message/19:77d621b48fd54d03ba7155385d33cb54@thread.skype/1596338973352?tenantId=4757b031-c3cc-4b2a-ac2c-a2ef4731ed22&groupId=7f2d402f-bd5b-4f3d-875f-fb78e2889c1a&parentMessageId=1596338973352&teamName=Omou&channelName=Defects&createdTime=1596338973352

jsflca2012:
Also fixed the error warning whenever invalid information is inputted into the text field


cypress/integration/Authentication/emailCheck_spec.js
```diff
@@ -0,0 +1,33 @@
...
+    first_name: "fake",
+    id: 2,
+    is_staff: true,
+    last_name: "person",
+};
+
+describe("emailCheck_spec", () => {
+    it("loads email check page", () => {
+        cy.visit("");
+        cy.get("[data-cy=nextButton]");
```
clam1987, line 20: this should contain an assertion to check against the button, not just grabbing the button data.

```
cy.get("[data-cy=nextButton]").should("contain", "SIGN IN");
```

cypress/integration/Authentication/emailCheck_spec.js
```diff
@@ -0,0 +1,33 @@
...
+        cy.get("[data-cy=createAccountButton]")
```
clam1987, line 21: same comment as above.

cypress/integration/Authentication/emailCheck_spec.js
```diff
@@ -0,0 +1,33 @@
...
+    });
+    it("checks invalid user", () => {
+        cy.get("[data-cy=emailField]").fastType(fakeDetails.email);
```
clam1987, line 24: you should make an assertion here to check if the correct email gets typed
```
 cy.get("[data-cy=emailField]").fastType(fakeDetails.email).should("have.value", fakeDetails.email);
```

cypress/integration/Authentication/emailCheck_spec.js
```diff
@@ -0,0 +1,33 @@
...
+        cy.get("[data-cy=nextButton]").click();
+        cy.get(".MuiFormHelperText-root").contains("Sorry, we couldn't find a user for that email")
```
clam1987, line 26: Don't use Mui class here, if MUI ever decides to change their class it'll break this test. Find where .MuiFormHelperText-root exists using react's dev tools and add a data-cy attribute to the element. If it does not exist, add it to the parent class and traverse the Virtual DOM to find it. Also, same comment as above. Instead of using contains, use should assertion.

cypress/integration/Authentication/emailCheck_spec.js
```diff
@@ -0,0 +1,33 @@
...
+    });
+    it("checks valid user", () => {
+        cy.get("[data-cy=emailField]").clear();
```
clam1987, line 29: Write an assertion after this line to make sure the input is cleared.

cypress/integration/Authentication/emailCheck_spec.js
```diff
@@ -0,0 +1,33 @@
...
+        cy.get("[data-cy=emailField]").fastType(userDetails.email);
```
clam1987, line 30: Same as above, make an assertion to check your input is correct.

cypress/integration/Authentication/emailCheck_spec.js
```diff
@@ -0,0 +1,33 @@
...
+        cy.get("[data-cy=nextButton]").click();
```
clam1987, line 31: Make an assertion afterwards to see if the password field appears.

cypress/integration/Authentication/emailCheck_spec.js
NeilAllavarpu: This might be a little too specific of a check, as earlier checks to make sure that the text field works already cover this case

clam1987 (dismissed):
Looks good to me!

clam1987 (changes requested):
Add the userDetails to user.json and you should be good to go

PR #553, opened 2020-08-26
Logout after 20 minutes idleness

rgedwill:
After 18 minutes, asks 'Are you still there?' If after 2 minutes, the confirmation isn't clicked, it automatically logs the user out.

Still need to write the tests for it and refactor it into its own component before it's ready for review.


dhuangg:
@rgedwill to confirm, this only prompts you with the "are you still there" model if you haven't touched anything for 20 minutes right? Meaning, if i were to click on something, the timer resets.

src/components/OmouComponents/IdleLogout.js
```diff
@@ -0,0 +1,178 @@
...
+
+    async function handleOpen() {
+        setOpen(true);
+        handleReset();
+        setRemaining(modalTimeout);
+        await logoutAfter2Minutes();
+
+    };
+    
+    const handleClose = () => {
```
dhuangg, line 109: use implicit returns `const handleClose = () => setOpen(false);`

src/components/OmouComponents/IdleLogout.js
```diff
@@ -0,0 +1,178 @@
...
+        handleClose();
+        handleLogout();
+    }
+
+    async function handleOpen() {
+        setOpen(true);
+        handleReset();
+        setRemaining(modalTimeout);
+        await logoutAfter2Minutes();
+
```
dhuangg, line 106: delete new line

src/components/OmouComponents/IdleLogout.js
```diff
@@ -0,0 +1,178 @@
...
+    };
+    
+    const handleClose = () => {
+        setOpen(false);
+    };
+
+    const {
+        reset,
+        pause,
+        resume,
```
dhuangg, line 116: what do "reset" "pause" and "resume" refer to? the timer?

src/components/OmouComponents/IdleLogout.js
```diff
@@ -0,0 +1,178 @@
...
+    
+        setInterval(() => {
+            setRemaining(getRemainingTime())
+            setLastActive(getLastActiveTime())
+            setElapsed(getElapsedTime())
+        }, 1000);
+    }, []);
+
+    const body = () => {
+        return (
```
dhuangg, line 143: If this is a component, capitalize the name and use implicit returns `()=> (<div> ... </div>)`

Be more descriptive than "body" what is this supposed to show?

src/components/OmouComponents/IdleLogout.js
```diff
@@ -0,0 +1,178 @@
...
+                        >Yes</Button>
+                    </div>
+                </p>
+            </div>
+        )};
+    return (
+        <div>
+            {
+                (remaining === 0) && handleOpen()
+            }
```
dhuangg, line 163: change 161 - 163 to 1 line

dhuangg: does "remaining" mean remaining seconds? milliseconds? make the variable name more specific

dhuangg (changes requested):
overall logic looks sound - there's some things to clean up. 

On your test, be sure to name the actual test instead of "Fills out form" which is a copy paste from Neil's test.

Also remember to resolve merge conflicts you committed as well as a current merge conflict with dev

src/components/OmouComponents/IdleLogout.js
```diff
@@ -0,0 +1,178 @@
...
+        width: '80%',
+        background: '#43b5d9',
+        border: '1px solid #43b5d9',
+        boxSizing: 'border-box',
+        boxShadow: '0px 1px 1px rgba(0, 0, 0, 0.25)',
+        borderRadius: '50px',
+        padding: '8px 0 8px 0',
+        color: 'white',
+        marginTop: '20px',
+
```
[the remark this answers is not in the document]
rgedwill: What white space is this referring to? Is it too big of an indent? Been having weird problems with my indents for js files in vscode, have to look into a fix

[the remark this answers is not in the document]
dhuangg: line 56 is empty

src/components/OmouComponents/IdleLogout.js
dhuangg: shouldn't this function actually refresh the token? I don't think this code actually refreshes token so even if you close this, the token will expire on the backend

dhuangg (changes requested):
took a second look at this. The concept of stubbing is to create fake API responses for each page we visit. 

What you copied over is Neil's stubs for the create/update admin form. What that means is that he had to mock both the "GetAdmin" query to pre-populate the form as well as the "CreateAdmin" mutation to update the admin. 

You're just testing the session logout functionality. To do so, you should: 
1. Go to the root URL (i.e. cy.visitAuthenticated("/") ) 
2. Mock the query on the dashboard view - "DashboardQuery" Dashboard.js line 49
3. Mock any other queries that you're calling (you should be calling the refresh token query/mutation, but I don't see you calling it right now in your component code).

src/components/OmouComponents/IdleLogout.js
```diff
@@ -53,98 +54,132 @@ const useStyles = makeStyles((theme) => ({
...
     const handleReset = () => reset();
-    const handlePause = () => pause();
-    const handleResume = () => resume();
-
-    useEffect(() => {
-        setRemaining(getRemainingTime())
-        setLastActive(getLastActiveTime())
-        setElapsed(getElapsedTime())
     
+    useEffect(() => {
```
dhuangg, line 171: I'm really confused what this useEffect does. First it sets the remaining miliseconds until "prompt" to whatever is "get remaining time." Then it sets the interval (of what?) to 1000 ms before it sets remaining miliseconds until prompt to whatever is "get remaining time."

I THINK it's counting down before the session logout prompt, then it updates the countdown every second?  Why is this necessary? 

Refactor the function names to make this more readable.

dhuangg (changes requested):
I'm not sure if this works or not, but overall, I'm just very confused by the code. take a look at my comments and let me know if this makes sense.

src/components/OmouComponents/IdleLogout.js
```diff
@@ -53,98 +54,132 @@ const useStyles = makeStyles((theme) => ({
...
+                        }
+                    });
+                });
+            }, 900000);
+        });
+    };
+
+    useEffect(() => {
+        (async () => {
+            await refreshTokenAfter15Minutes();
```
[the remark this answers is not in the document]
rgedwill: It should only run once when the component renders initially - I needed to add an empty dependency array. Just pushed

src/components/OmouComponents/IdleLogout.js
rgedwill: The first line in that function was unecessary - I forgot to delete when playing with the token. 

The setInterval is what actually tracks the countdown - getRemainingTime is the react-idle-timer call. It's at an interval so it isn't being called more than every second since that's as precise as necessary to update the state.

dhuangg: what's the difference between this useEffect and the earlier useEffect (line 135)?

dhuangg (changes requested):
overall I'm still very confused as to what and how does this component flow. do you have a process flow diagram of how this is supposed to work?

rgedwill:
> overall I'm still very confused as to what and how does this component flow. do you have a process flow diagram of how this is supposed to work?

I'll make a diagram real quick

src/components/OmouComponents/IdleLogout.js
```diff
@@ -1,3 +1,13 @@
+/*
+    This token handles 2 things.
```
dhuangg, line 2: what token?

src/components/OmouComponents/IdleLogout.js
```diff
@@ -117,6 +131,7 @@ const IdleLogout = () => {
         })
     };
     
+    // Refreshes the token every 15 minutes
     const refreshTokenAfter15Minutes = () => {
```
dhuangg, line 135: The comment here isn't very helpful - it's just what you're saying in the function name. Why are you refreshing every 15 minutes?

src/components/OmouComponents/IdleLogout.js
```diff
@@ -1,3 +1,13 @@
+/*
+    This token handles 2 things.
+    1. Refreshes the authentication token every 15 minutes while the user is loggeed in. This 
+        ensures a logged in user always has a (relatively) fresh token
```
dhuangg, line 4: Why are we refreshing the auth token every 15 minutes?

src/components/OmouComponents/IdleLogout.js
```diff
@@ -82,12 +90,16 @@ const IdleLogout = () => {
     // Time(ms) user has to click that they're still here before they're logged out
     const modalTimeout = 180000;
 
+    // State variable being updated by the idle timer - when this variable reaches 0 the
+    // modal will display
     const [remainingMsUntilPrompt, setRemainingMsUntilPrompt] = useState(idleTimeout);
```
dhuangg, line 95: Explain what is an idle timer - i'm assuming this is the timer that's tracking the time after the last time a user has performed something

src/components/OmouComponents/IdleLogout.js
```diff
@@ -131,44 +146,37 @@ const IdleLogout = () => {
...
     async function handleOpen() {
         setOpenModal(true);
-        handleReset();
-        setRemainingMsUntilPrompt(modalTimeout);
-
-        sessionTimeoutRef.current = setTimeout(logoutAndCloseModal, 5 * 1000);
+        sessionTimeoutRef.current = setTimeout(logoutAndCloseModal, modalTimeout);
     }
     
+    // Close the modal, clear the modal timeout ref, reset the token
```
dhuangg, line 162: use comments to explain why you're doing something - it's clear from the function that you're clearInterval(), setOpenModal(false) and resetToken(), but it's not clear why and when this function is used.

src/components/OmouComponents/IdleLogout.js
```diff
@@ -131,44 +146,37 @@ const IdleLogout = () => {
...
     const {
-        reset,
         getRemainingTime,
     } = useIdleTimer({
         "timeout": idleTimeout
     });
     
-    const handleReset = () => reset();
-    
+    // Resets token every 15 minutes and sets the intervals for the idle-timeout function
```
dhuangg, line 175: Be descriptive as to why you're doing this - my understanding is that you're doing this to kick off the 15 minute on component load so it will keep refreshing the token every 15 minutes as long as the idleTimer never hits 18 minutes?

dhuangg (changes requested):
This is better and the code makes more sense. But focus on explaining the logic/why you chose this session timeout logic. Some of these comments are just restating what is evident in your code which is a bit redundant.

src/components/OmouComponents/IdleLogout.js
```diff
@@ -85,18 +75,18 @@ const IdleLogout = () => {
     }, [dispatch, history]);
     
     // Time(ms) before modal pops up
-    const idleTimeout = 1080000; 
+    const idleTimeout = 5000; //1080000;
```
dhuangg, line 78: isn't this 5 seconds?

dhuangg (changes requested):
looks good - just change the timeout time back to 18 minutes

PR #555, opened 2020-09-11
Fix for Issue A027 (Autocomplete No Options Text) and A045 (Phone Number regex)

andrewjknapp:
Added noOptionsText prop to Fields.DataSelect components within parentSelect, categorySelect, and schoolSelect functions in FormFormats.js. Also put {...props} on its own line in Fields.js line 79.

### DIFF
--- a/src/components/Form/FormFormats.js
+++ b/src/components/Form/FormFormats.js
@@ -69,7 +69,7 @@ const userMap = ({accountSearch}) => accountSearch.results.map(({user}) => ({
 
 const instructorSelect = (name) => (
     <Fields.DataSelect name={name} optionsMap={userMap}
-        request={SEARCH_INSTRUCTORS} />
+        request={SEARCH_INSTRUCTORS} noOptionsText={"No instructors available"}/>
 );
 
 
--- a/src/components/Form/FormFormats.js
+++ b/src/components/Form/FormFormats.js
@@ -68,8 +68,10 @@ const userMap = ({accountSearch}) => accountSearch.results.map(({user}) => ({
 }));
 
 const instructorSelect = (name) => (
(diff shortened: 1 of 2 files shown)


dhuangg:
@jsflca2012 please review this when you can. I'll review after you

jsflca2012 (commented):
Looks good, just remove console.log here

andrewjknapp:
Thank you for catching that. I removed the console.log and the curly braces around the noOptionsText.

src/components/Form/FormFormats.js
```diff
@@ -69,7 +69,7 @@ const userMap = ({accountSearch}) => accountSearch.results.map(({user}) => ({
 
 const instructorSelect = (name) => (
     <Fields.DataSelect name={name} optionsMap={userMap}
-        request={SEARCH_INSTRUCTORS} noOptionsText={"No instructors available"}/>
+        request={SEARCH_INSTRUCTORS} noOptionsText="No instructors available"/>
```
dhuangg, line 72: each attribute should be its own line. example:
```
<Fields.DataSelect name="name"
                   request={REQUEST}
                   noOptionsText="No options"
/> 
```

dhuangg: This should be applied to everything

dhuangg (changes requested):
Good changes - apply the last syntax change and we can merge this in

src/components/Form/FormFormats.js
andrewjknapp: Ok, added that formatting. Thanks

dhuangg (approved):
looks great! congrats on your first PR!

PR #560, opened 2020-09-18
Created a reusable AddItemButton component

pswk1:
Created a AddItemButton.js in FeatureViews > Accounts > TabComponents. Takes in height & width props to determine its size depending on where it's being used. For example, it's used in StudentInfo.js in place of the Link that was placed in the Grid item. In this case a component and to prop are being passed for its function, but it can also take some onClick function as well.


src/components/FeatureViews/Accounts/ProfileCard.js
```diff
@@ -50,6 +50,8 @@ const ProfileCard = ({ user, route, studentInvite = false }) => {
         });
     }, [invite, user]);
 
+    console.log(user);
```
dhuangg, line 53: remove console log

src/components/FeatureViews/Accounts/TabComponents/AddItemButton.js
```diff
@@ -0,0 +1,45 @@
...
+    <Box {...rest}>
+      <div className={addNewItemStyles}>
+        {children}
+      </div>
+    </Box>
+  );
+};
+
+AddItemButton.propTypes={
+  height: PropTypes.oneOfType([
```
dhuangg, line 37: make sure the height is required

dhuangg (changes requested):
Overall looks good! Just write a cypress test to validate that the component properly sizes the height when you give it a valid height value. 

Also move this component into the "OmouComponents" directory

rgedwill:
@dhuangg should I wait for the cypress tests to review?

dhuangg:
go ahead and leave your review as well. if you can't find anything then give peter a poke haha

src/components/FeatureViews/Accounts/TabComponents/AddItemButton.js
```diff
@@ -0,0 +1,45 @@
...
+    width: width,
+    background: '#F5F5F5',
+    border: '1.5px dashed #999999',
+    borderRadius: '5px',
+    boxSizing: 'border-box',
+    boxShadow: '0 0.5rem 0.5rem rgba(0,0,0, 0.2)',
+  }),
+});
+
+export const AddItemButton = ({ children, ...rest }) => {
```
rgedwill, line 24: Can you make a variable name more descriptive than 'rest'?

src/components/FeatureViews/TeachingLog/TeachingLogContainer.js
```diff
@@ -279,7 +279,10 @@ export default function TeachingLogContainer() {
 									<b>Total Hours</b>
 								</TableCell>
 								<TableCell>
-									{sessions.length > 0 && summaryLog.reduce((acc, course) => (acc.hours + course.hours))}
+									{sessions.length > 0 && 
+									summaryLog.reduce((acc, course) => (acc.hours + course.hours)) === Number.NaN
```
rgedwill, line 283: Since this error isn't too descriptive anyways, you could probably make this more readable and DRY by using optional chaining. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

rgedwill (approved):
Overall looks very good, could potentially use an improvement or 2 but they're just suggestions not requirements

src/components/FeatureViews/TeachingLog/TeachingLogContainer.js
pswk1: `{sessions.length > 0 ? summaryLog.reduce((acc, course) => (acc?.hours + course?.hours)) : "Error"`
Just to check for understanding, if the hours didn't exist in acc or course, that reduce would return undefined due to optional chaining instead of NaN, and thus would still display "Error", correct?

cypress/integration/Add_Item_Button.js
```diff
@@ -0,0 +1,7 @@
+describe("Properly calculates height of AddItemButton component", () => {
+    it('properly reads the height prop of AddItemButton component and sets it accordingly ', () => {
+        cy.visit('/AddItemButtonTest');
+        cy.get('.makeStyles-addNewItemStyles-3').should('have.css', 'height', '240px');
```
dhuangg, line 4: don't use css classes to select for components. use the data-cy attribute. 

Reference: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements

src/components/Routes/RootRoutes.js
```diff
@@ -60,6 +62,12 @@ export const RootRoutes = () => {
                 <NewAccount/>
             </Route>
 
+            {/* Route for Testing AddItemButton */}
+            <Route path="/AddItemButtonTest">
```
dhuangg, line 66: For any demo views, let's have it route to `/demo/[componentName]` in this case, this should be `/demo/AddItemButton`

dhuangg (changes requested):
small changes, almost done!

dhuangg (approved):
looks good - make sure to remove the curly bracket!

PR #562, opened 2020-09-19
Fixed account search filter

juicekey:
Profile from search query needed to be uppercase (#461). Changed menuItem value for receptionist (AccountFilters.js).


.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
dhuangg, line 1: be sure to gitignore this file

src/components/FeatureViews/Search/SearchResults.js
```diff
@@ -83,13 +83,13 @@ const SearchResults = () => {
     const filter = searchParams.get("filter"),
         query = searchParams.get("query"),
         sort = searchParams.get("sort");
-
+    const profile = searchParams.get("profile")?.toUpperCase()
```
dhuangg, line 86: small syntax improvement - continue off of the const variable definitions started on line 83. you can delete the semi-colon on line 85 and add a comma to define another const variable

dhuangg (changes requested):
this looks good! just fix the small syntax improvement comment and add your vscode project file to the .gitignore so we don't add this to the repo

dhuangg (approved):
awesome work!

PR #563, opened 2020-09-20
Display "..." when email is longer than 21 characters

juicekey:
Added conditional rendering of "..." after existing 21 character substring.


src/components/FeatureViews/Accounts/Accounts.js
```diff
@@ -198,7 +198,10 @@ const Accounts = () => {
                         </TableCell>
                         <TableCell>
                             <Tooltip title={row.user.email}>
-                                <span>{row.user.email.substr(0, 20)}</span>
+                                <span>
+                                    {row.user.email.substr(0, 20)}
+                                    {(row.user.email.length > 21) && "..."}
```
dhuangg, line 203: 2 things:
1. avoid having magic numbers - define them in a const variables (remember to use all caps, example: MAX_EMAIL_LENGTH) 
2. use an explainer variable for your condition. could be something like: `isOverMaxEmailLength`

dhuangg (changes requested):
also looks pretty good, just small syntax/clean code comments!

src/components/FeatureViews/Accounts/Accounts.js
```diff
@@ -162,6 +162,11 @@ const Accounts = () => {
         setViewToggle(view);
     }, []);
 
+    const MAX_EMAIL_LENGTH = 21;
+    const isOverMaxEmailLength = (emailLength) => {
+        return emailLength > MAX_EMAIL_LENGTH
```
dhuangg, line 167: use implicit return statement

dhuangg (approved):
change small syntax thing and you're good to merge!

PR #565, opened 2020-09-24
Added sort feature to Registration Landing

juicekey:
Added a dropdown menu allowing for sort by course title and remaining seats.

### DIFF
--- a/src/components/FeatureViews/Registration/RegistrationLanding.js
+++ b/src/components/FeatureViews/Registration/RegistrationLanding.js
@@ -84,6 +84,7 @@ const RegistrationLanding = () => {
         "instructor": [],
         "subject": [],
     });
+    const [sortType, setSortType] = useState(null);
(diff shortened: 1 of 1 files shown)


src/components/FeatureViews/Registration/RegistrationLanding.js
```diff
@@ -145,6 +145,23 @@ const RegistrationLanding = () => {
                     return courses;
             }
         }, Object.values(courses));
+    
+    const sortedCourses = filteredCourses.sort((a,b) => {
```
CalvinFronda, line 149: Change the variables "a" and "b" to something more meaningful.

CalvinFronda (changes requested):
value is not being updated so if you change the filter, the input just shows the place holder.

dhuangg (approved):
looks good! 🥇

PR #567, opened 2020-09-25
Bm parent instructor view

clam1987:


cypress/integration/User Management/BM_00_ParentInstructorView.js
```diff
@@ -0,0 +1,60 @@
...
+            cy.mockGraphQL({
+                "getCourses": {
+                    "response": {
+                        "data": {
+                            "courses": AllCourses,
+                        },
+                    },
+                    "test": (x) => {
+                        // console.log(x)
+                        // expect(id).equals(parent.user.id.toString(), "Check ID passed");
```
dhuangg, line 13: remove test key if you don't have any tests

cypress/integration/User Management/BM_00_ParentInstructorView.js
```diff
@@ -0,0 +1,60 @@
+describe("Test should load our courses from course.json into course management, check to see if parent's student is enrolled in a course, then view instructor account", () => {
```
dhuangg, line 1: This should be a high level description of the entire test flow so you shouldn't include "course.json". I'm honestly not sure what you're testing for here. Are you testing your course management flows from the parent's portal? don't combine both parents and instructor portal tests, these should be separate tests.

cypress/integration/User Management/BM_00_ParentInstructorView.js
```diff
@@ -0,0 +1,60 @@
...
+            cy.fixture("announcements.json").then(({ announcements }) => {
+                cy.mockGraphQL({
+                    "getAnnouncements": {
+                        "response": {
+                            "data": announcements,
+                        },
+                        "test": (x) => {
+                            console.log(x)
+                            // expect(id).equals(parent.user.id.toString(), "Check ID passed");
+                        },
```
dhuangg, line 50: remove if you're not using

cypress/integration/User Management/BM_00_ParentInstructorView.js
```diff
@@ -0,0 +1,60 @@
...
+                    .should("have.css", "color", "rgb(40, 171, 213)");
+                cy.get("[data-cy=course-list-chip]")
+                    .last()
+                    .should("have.text", "PAST")
+                    .should("have.css", "background-color", "rgb(189, 189, 189)");
+        });
+    });
+
+    it("Will click on the first course, taking the parent to the course information page", () => {
+        before(() => {
```
dhuangg, line 40: This `before()` shouldn't be in a `it()` function. it should be in the `before()`function at the top

dhuangg (changes requested):
This is a work in progress but take a look at my comments

Issue #559, opened 2020-09-17
Parse phone number input from phone number field


dhuangg:
Currently our regex for our phone input accepts a lot of different versions (e.g. (xxx) xxx-xxxx, xxx-xxx-xxxx, etc.) we need to parse these out to this form: xxxxxxxxxx so that we can display all phone numbers consistently. 

Potentially look into integrating this [react-phone-number-input](https://www.npmjs.com/package/react-phone-number-input) package. 
- [Stack overflow question on potentially how to implement this](https://stackoverflow.com/questions/55687357/how-to-add-react-phone-number-input-to-react-final-form)
- [Final Form Field documentation](https://final-form.org/docs/react-final-form/api/Field)

andrewjknapp:
react-phone-number-input seems very useful. One of the options with this package is to allow the user to choose which country their phone number is registered in. Should I include this functionality or are we assuming that all users will has a US phone number?

PR #571, opened 2020-09-29
Fix for #559 Added Text mask to Phone Input

andrewjknapp:
Added a text mask to the phone input fields so now when the user enters numbers the form automatically wraps the numbers into this format (111) 222-3333 as they type. This is a fix for #559


dhuangg (approved):
looks good!

PR #579, opened 2020-10-05
Action log

tigermli:
Upon testing it was found by @jsflca2012 that the Action Log object filter was broken. The following fixes should resolve this.
For ActionLog object type filter:
- Converted "Price Rules" to "pricerule"
- Removed "Tutoring" from filter


dhuangg (approved):
looks good!

PR #587, opened 2020-10-10
Rich text editor

clam1987:
Working on refactoring some components to make it more readable


src/components/FeatureViews/Courses/Announcements.js
```diff
@@ -56,17 +64,17 @@ const AnnouncementCard = ({
   handleEdit,
   handleDelete,
 }) => {
+  const decode = EditorState.createWithContent(convertFromRaw(JSON.parse(body)))
```
NeilAllavarpu, line 67: Why do you have to `JSON.parse` `body`? Is it not possible to pass in the JSON directly?

src/components/FeatureViews/Courses/ModalTextEditor.js
```diff
@@ -134,122 +185,309 @@ const ModalTextEditor = ({
...
+  textSubject,
+  textBody,
+  buttonState,
+  announcementId,
+}) => {
+  const classes = useStyles();
+  const [sendEmailCheckbox, setSendEmailCheckbox] = useState(false);
+  const [sendSMSCheckbox, setSendSMSCheckbox] = useState(false);
+  const [subject, setSubject] = useState(textSubject);
+  const [body, setBody] = useState(() => EditorState.createEmpty());
```
NeilAllavarpu, line 218: Optional: since you call this function on its own and return its value directly, the arrow function is not necessary; you can simply put `useState(EditorState.createEmpty)`, and the function is called as expected.

src/components/FeatureViews/Courses/ModalTextEditor.js
```diff
@@ -134,122 +185,309 @@ const ModalTextEditor = ({
...
     } else {
-      const createdSessionNote = await createSessionNote({
-        variables: {
-          subject: subject,
-          body: body,
-          user: poster_id,
-          sessionId: sessionId,
-          id: noteId,
+      setBody(EditorState.createEmpty())
+    }
```
NeilAllavarpu, line 227: Do we plan to have more than two button states? If so, a switch statement might be more appropriate; if not, consider if a boolean variable would be more approriate (perhaps `isEditing`).

src/components/FeatureViews/Courses/ModalTextEditor.js
```diff
@@ -134,122 +185,309 @@ const ModalTextEditor = ({
...
+    });
+    setBody(EditorState.createEmpty())
+  };
+
+  const toggleInlineStyles = e => {
+    e.preventDefault();
+    const style = e.currentTarget.getAttribute("data-style");
+    const { toggleInlineStyle, toggleBlockType } = RichUtils
+    const newState = {
+      [style]: toggleInlineStyle(body, style),
```
NeilAllavarpu, line 314: As written, this is the only key of the object that is ever used; did you mean to use quotes instead of brackets?

src/components/FeatureViews/Courses/ModalTextEditor.js
```diff
@@ -134,122 +185,309 @@ const ModalTextEditor = ({
...
+      "unordered-list-item": toggleBlockType(body, style),
+      "ordered-list-item": toggleBlockType(body, style),
+    }; 
+    if(newState[style]) {
+      setBody(newState[style])
+      return "handled"
+    }
+    return "not handled"
```
NeilAllavarpu, line 322: What's the significance of the return statement? If it's not used, remove it.

src/components/FeatureViews/Courses/ModalTextEditor.js
```diff
@@ -134,122 +185,309 @@ const ModalTextEditor = ({
...
+  };
+
+  const handleKeyCommand = (command, body) => {
+    const newState = RichUtils.handleKeyCommand(body, command);
+    if (newState) {
+      setBody(newState);
+      return "handled";
     }
+    return "not handled";
```
NeilAllavarpu, line 331: See above

src/components/FeatureViews/Courses/ModalTextEditor.js
```diff
@@ -134,122 +185,309 @@ const ModalTextEditor = ({
...
+  const renderButtonText = () => {
+    switch (buttonState) {
+      case "edit":
+        return "EDIT";
+      case "post":
+        return "POST";
+      default:
+        return "ADD NOTE";
+    }
+  };
```
NeilAllavarpu, line 354: Optional: since you only want static values, a lookup table might be more appropriate, e.g.

    const buttonTextParser = {
        "edit": EDIT,
        ...

using a default value if the key does not exist. This may or may not be more complex, depending on the code structure; check for yourself if it's worth converting.

Also, since you only plan to compute this once whenever buttonState is a certain value, a computed value would be more appropriate; call the function, store the value into a constant (perhaps via `useMemo`) and then used that constant, for both simplicity and a bit of optimization.

NeilAllavarpu (changes requested):

src/components/FeatureViews/Courses/Announcements.js
clam1987: When the body is loaded from the database, it's loaded as a string so I'd have to parse it back to an object.

src/components/FeatureViews/Courses/ModalTextEditor.js
```diff
@@ -66,190 +79,261 @@ const useStyles = makeStyles((theme) => ({
...
 }) => {
   const classes = useStyles();
-  const [subject, setSubject] = useState(noteSubject);
-  const [body, setBody] = useState(noteBody);
-  const poster_id = posterId.results[0].user.id;
+  const [sendEmailCheckbox, setSendEmailCheckbox] = useState(false);
+  const [sendSMSCheckbox, setSendSMSCheckbox] = useState(false);
+  const [subject, setSubject] = useState();
+  const [body, setBody] = useState(EditorState.createEmpty());
+  const isDisabled = convertToRaw(body.getCurrentContent())?.blocks
```
dhuangg, line 109: is what disabled?

src/components/FeatureViews/Courses/ModalTextEditor.js
```diff
@@ -66,190 +79,261 @@ const useStyles = makeStyles((theme) => ({
...
-          id: noteId,
-        },
-      });
+      setBody(toggleInlineStyle(body, style))
     }
   };
 
-  const handleSubjectChange = useCallback((event) => {
+  const handleKeyCommand = (command, body) => {
+    const newState = RichUtils.handleKeyCommand(body, command);
```
dhuangg, line 192: newTextState?

src/utils.js
```diff
@@ -607,3 +607,20 @@ export function useSessionStorage(key, initialValue) {
 
     return [storedValue, setValue];
 }
+
+/** 
+ * @description
```
dhuangg, line 612: write documentation of when to use this function. what should be expected of "newData" and "cachedData" (i.e. what should the data types be for these params?)

dhuangg (changes requested):
sorry, long overdue review, but just some small things to change. rest looks good. 

may take some time to merge everything

Issue #490, opened 2020-08-04
A063 - Can't look at course info as parent


dhuangg:
A063
Can't look at course info as parent
Bug found by: Ryan
Teams link: https://teams.microsoft.com/l/message/19:77d621b48fd54d03ba7155385d33cb54@thread.skype/1596407665087?tenantId=4757b031-c3cc-4b2a-ac2c-a2ef4731ed22&groupId=7f2d402f-bd5b-4f3d-875f-fb78e2889c1a&parentMessageId=1596407665087&teamName=Omou&channelName=Defects&createdTime=1596407665087

andrewjknapp:
This is what the Parent view can look like. I surrounded the Tabs component with an AccessControlComponent which is only visible to admin, receptionist, and instructors. This also hides the notes tab, is this something that we want the parents to be able to see? If so I could use the AccessControlComponent to only hide the Tab which contains the information about all enrolled students.
![image](https://user-images.githubusercontent.com/16299570/93692648-a7833680-faaa-11ea-9ed7-9d073c9e6b99.png)

PR #593, opened 2020-10-13
A063 parent cannot see course info

andrewjknapp:
Parents were previously unable to view the courses under Registration. Now they have access to see the course info and the information that they should not see, such as the list of enrolled students, is hidden with an AccessControlComponent. This is a fix for #490


clam1987 (approved):
Everything looks good! Great job on this Andrew!

dhuangg (approved):
👍

PR #611, opened 2020-11-05
Refactor payment table

andrewjknapp:


src/components/FeatureViews/Accounts/TabComponents/EnrollmentPayment.js
```diff
@@ -0,0 +1,69 @@
...
+import PaymentTable from "./PaymentTable";
+import {useQuery} from "@apollo/react-hooks";
+import Typography from "@material-ui/core/Typography";
+import gql from "graphql-tag";
+
+export const GET_ENROLLMENT_PAYMENTS = gql`
+query EnrollmentPayments($enrollmentId: ID!) {
+  enrollment(enrollmentId: $enrollmentId) {
+    paymentList {
+      __typename
```
CalvinFronda, line 13: You don't need to pass ```__typename``` if you aren't using it. They automatically populate after you make a graphql request.

src/components/FeatureViews/Accounts/TabComponents/EnrollmentPayment.js
```diff
@@ -0,0 +1,69 @@
...
+        }
+        id
+        numSessions
+        updatedAt
+      }
+      enrollments {
+        id
+      }
+    }
+    __typename
```
CalvinFronda, line 37: same as above

src/components/FeatureViews/Accounts/TabComponents/EnrollmentPayment.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+`;
+
+const EnrollmentPayment = ({enrollmentID, courseID}) => {
+	const {data, loading, error} = useQuery(GET_ENROLLMENT_PAYMENTS,
+		{variables: {enrollmentId: enrollmentID}}
+	);
+    
+	if (loading) {
```
CalvinFronda, line 47: Not really an issue but it makes your code less lines - you use conditionals without the brackets. 

```
if (loading) return <Loading/>
if (error) return <Typography>There's been an error! - {error.message}</Typography>
```

src/utils.js
```diff
@@ -218,7 +218,7 @@ export const paymentToString = (string) => {
         case "credit_card":
             return "Credit Card";
         default:
-            return capitalizeString(string);
+            return capitalizeString(string.toLowerCase());
```
CalvinFronda, line 221: Did some digging and it looks like some of the queries might be using the old REST responses. Can you change the ```paymentToString``` function to take in the new graphQL responses? You can just copy and past the code below. 
```
export const paymentToString = (string) => {
    switch (string) {
        case "INTL_CREDIT_CARD":   
            return "International Credit Card"; 
        case "CREDIT_CARD":
            return "Credit Card";
        default:
            return capitalizeString(string);
    }
};
```

CalvinFronda (changes requested):
Overall good work. I would say take a look at the query again and only use what you need. It looks like you are only using ```id```, ```createdAt```, ```total```, ```registrationSet``` and ```method```

src/components/FeatureViews/Accounts/TabComponents/EnrollmentPayment.js
andrewjknapp: Ok Great, made that update

src/components/FeatureViews/Accounts/TabComponents/EnrollmentPayment.js
andrewjknapp: Ok Great, made that update

src/components/FeatureViews/Accounts/TabComponents/EnrollmentPayment.js
andrewjknapp: Ok Great, made that update

src/utils.js
andrewjknapp: I can do that, but are there pages that still get the REST responses that would break if we change this? Should I go through and check where this is used to see if changing it would cause problems?

clam1987 (approved):
Looks good!

CalvinFronda (commented):
looks good.

PR #635, opened 2020-11-27
Dahl accounts card

pswk1:
Refactored ProfileCard.js to match specifications in Dahl Figma, as well as AccountCard.js that is displayed when a parent is set in Registration


src/theme/ThemedComponents/AccountCard/AccountCard.js
```diff
@@ -0,0 +1,121 @@
...
+	}
+});
+
+const AccountCard = ({ accountName, role }) => {
+	const classes = useStyles();
+
+	let stripeColor;
+
+	if (role === 'Parent') {
+		stripeColor = '#D74AC8';
```
andrewjknapp, line 58: Change this to reflect how the color is determined with the stringToColor function

src/components/FeatureViews/Accounts/ProfileCard.js
```diff
@@ -53,68 +98,75 @@ const ProfileCard = ({ user, route, studentInvite = false }) => {
     }, [invite, user]);
 
     const classes = useStyles();
+    const stripeColor = stringToColor(user.name);
```
andrewjknapp, line 101: The only place you use this is on line 109. You could instead just call the function there and get rid of this variable.

andrewjknapp (changes requested):
A few minor changes I have listed. Also, how different are the Account Cards and the Profile Cards? Would it make sense to make a single component and use them in both locations? Let me know what you think.

andrewjknapp (approved):
Thank you for addressing all of the topics we discussed.

PR #637, opened 2020-12-01
Created list component and applied to app

juicekey:
Created reusable component for lists.

Updated lists in registration and course management to figma styling.

Course name is clickable, not entire component, in order to avoid routing issue when clicking instructor name.


andrewjknapp:
It looks there are a couple issues with the upper margin on the list in course management

image

As well as in Registration

image

dhuangg (approved):
This looks good! I like how you're handling the link clicking - you just didn't build that behavior into the component itself and used a <Box onClick/> event listener :)

andrewjknapp (commented):
Glad to see the margins are fixed. Good job on this one.

PR #638, opened 2020-12-01
Component viewer update

CalvinFronda:
Updated component viewer to GraphQL. Accounts work now


dhuangg (commented):
i put some comments on there for your specific changes, but there's a lot of changes that were rolled back because of this merge. I'm not sure if these are on dev, but double check on this

PR #639, opened 2020-12-01
Component viewer instructor migration

andrewjknapp:


src/components/FeatureViews/Accounts/TabComponents/InstructorSchedule.js
```diff
@@ -1,130 +1,124 @@
...
-				({day, start, end}) => ({
-					daysOfWeek: [day],
-					endTime: end,
-					startTime: start,
-				})
-				)
-				: [],
-		[instructor]
-	);
+	if (loading ) return null;
```
dhuangg, line 54: use the `<Loading small/>` component

src/components/FeatureViews/Accounts/TabComponents/InstructorCourses.js
```diff
@@ -1,56 +1,52 @@
...
-    if (hooks.isLoading(courseStatus)) {
-      return <Loading />;
-    }
-    if (hooks.isFail(courseStatus)) {
-      return <LoadingError error="courses" />;
-    }
-  }
+	if (loading ) return null;
+
+  if (error ) return "Unable to load courses";
```
dhuangg, line 39: There should be an <Error/> component as well, not sure, but double check

src/components/FeatureViews/Accounts/TabComponents/InstructorSchedule.js
```diff
@@ -1,130 +1,124 @@
...
-				classEnrollmentStatus,
-				availabilityStatus
-			)
-		) {
-			return "Unable to load schedule!";
+	const { user } = data.instructor;
+
+	const today = new Date(Date.now());
+
+	const hoursWorkedThisMonth = parseInt(user.instructor.sessionSet.reduce((hours, session) => {
```
dhuangg, line 62: does this work? like under the user object there's an instructor key? if this is the case then you may want to destructure line 58 to be like: `const { user: { instructor } } = data. instructor` or `const { instructor } = data.instructor.user`

src/components/FeatureViews/Accounts/TabComponents/InstructorSchedule.js
```diff
@@ -1,130 +1,124 @@
...
 
+	const instructorBusinessHours = user.instructor.instructoravailabilitySet.map(({dayOfWeek, startTime, endTime}) => ({
+		dayOfWeek: [dayOfWeek],
+		startTime, 
+		endTime
+	}))
+
+	const teachingSessions = user.instructor.sessionSet.map(({endDatetime, startDatetime, title}) => ({
+		title: title,
+		start: new Date(startDatetime),
```
dhuangg, line 86: use moment instead of `new Date`

src/components/FeatureViews/Accounts/TabComponents/InstructorSchedule.js
```diff
@@ -1,130 +1,124 @@
...
+		end: new Date(endDatetime)
+	}));
+
+	let allDayOOO = false;
+	// Out Of Office
+	const OOO = user.instructor.instructoroutofofficeSet.map(({startDatetime, endDatetime, description}) => {
```
dhuangg, line 92: rename to OOOEvents

src/components/FeatureViews/Accounts/TabComponents/InstructorSchedule.js
```diff
@@ -1,130 +1,124 @@
...
+		let start;
+		let allDay = false;
+		// Checks to see if the start and end time span the whole day
+		if (toHours(new Date(startDatetime) - new Date(endDatetime)) >= 24 ) {
```
dhuangg, line 96: use moment instead of new Date. moment has a .diff function you can use https://momentjs.com/docs/#/displaying/difference/

src/components/FeatureViews/Accounts/TabComponents/InstructorSchedule.js
```diff
@@ -1,130 +1,124 @@
...
+			start = null;
+			allDay = true;
+			allDayOOO = true;
+		} else {
+			start = new Date(startDatetime);
```
dhuangg, line 101: use moment

src/components/FeatureViews/Accounts/TabComponents/InstructorSchedule.js
```diff
@@ -1,130 +1,124 @@
...
-				tutoringEnrollmentStatus,
-				OOOstatus,
-				classEnrollmentStatus,
-				availabilityStatus
-			)
-		) {
-			return "Unable to load schedule!";
+	const { user } = data.instructor;
+
+	const today = new Date(Date.now());
```
dhuangg, line 60: use moment

dhuangg (changes requested):
looks mostly good - just need to use moment instead of the default javascript Date object :)

src/components/FeatureViews/Accounts/TabComponents/InstructorCourses.js
```diff
@@ -1,56 +1,52 @@
...
 
-  if (Object.keys(courses).length === 0) {
-    if (hooks.isLoading(courseStatus)) {
-      return <Loading />;
-    }
-    if (hooks.isFail(courseStatus)) {
-      return <LoadingError error="courses" />;
-    }
-  }
+	if (loading ) return null;
```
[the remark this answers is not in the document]
andrewjknapp: Ok updated this.

src/components/FeatureViews/Accounts/TabComponents/InstructorCourses.js
andrewjknapp: Replaced the string with <Error/>

src/components/FeatureViews/Accounts/TabComponents/InstructorSchedule.js
andrewjknapp: I need access to the user object, so I changed it to `const { user, user : { instructor } } = data.instructor; `

src/components/FeatureViews/Accounts/TabComponents/InstructorCourses.js
```diff
@@ -8,35 +8,53 @@ import Paper from "@material-ui/core/Paper";
...
+        id
+        dayOfWeek
+        endTime
+        startTime
+      }
     }
   }
 `;
 
+const checkTimes = (courseTimes) => {
```
dhuangg, line 37: What are you checking for? rename to clarify

dhuangg: This seems like a weird check - you're checking to see if there are any other start/end times?

src/components/FeatureViews/Accounts/TabComponents/InstructorCourses.js
```diff
@@ -8,35 +8,53 @@ import Paper from "@material-ui/core/Paper";
...
+  let start = courseTimes[0].startTime;
+  let end = courseTimes[0].endTime;
+
+  for (let course of courseTimes) {
```
dhuangg, line 41: this doesn't make sense - shouldn't it be `let availability of courseAvailabilities` ?

src/components/FeatureViews/Accounts/TabComponents/InstructorCourses.js
```diff
@@ -8,35 +8,53 @@ import Paper from "@material-ui/core/Paper";
...
 
   const { loading, error, data } = useQuery(GET_INSTRUCTOR_ENROLLMENTS, {
 		variables: {instructorID}
 	})
 
-	if (loading ) return null;
+	if (loading ) return <Loading small/>;
 
-  if (error ) return "Unable to load courses";
+  if ( error ) return console.error(error.message);
```
dhuangg, line 57: there should be an error component you can use

src/components/FeatureViews/Accounts/TabComponents/InstructorCourses.js
```diff
@@ -108,28 +123,31 @@ const InstructorCourses = ({ instructorID }) => {
                     </Grid>
                     <Grid item xs={2}>
                       <Typography align="left">
-                        <Moment
-                            format="dddd"
-                            date={startDate}
-                        />
+                        {availabilityList.map(({dayOfWeek}, index) => {
+                          let isLast = availabilityList.length - 1 === index;
```
dhuangg, line 127: this should be "isLastDayOfWeek"

dhuangg (changes requested):
small changes to names then we're done 👍

src/components/FeatureViews/Accounts/TabComponents/InstructorCourses.js
andrewjknapp: This check is to see if all times for the course are the same, like Wednesday at 4-5 and Thursday at 4-5. If so then the times will be displayed in the table. If they are different then the word "Various" will be displayed in the table instead. I can change the name to clarify the names

src/components/FeatureViews/Accounts/TabComponents/InstructorCourses.js
andrewjknapp: Good call, I'll change that.

src/components/FeatureViews/Accounts/TabComponents/InstructorCourses.js
andrewjknapp: I looked around and I could not find one. I saw in multiple other components doing this or` <Typography>There was an error {message.error}</Typography>`. Where can I find the <Error/>, is that part of Material UI?

dhuangg: hm... i thought Jeffrey made one. it's ok. but yes, you do want to display this on the component itself

andrewjknapp: Ok I do not see an Error component inside of OmouComponents, so I will do  `<Typography>There was an error {message.error}</Typography>` instead.

dhuangg:
looks good! @CalvinFronda should check before merge

src/components/FeatureViews/Accounts/TabComponents/InstructorSchedule.js
CalvinFronda: Change the ```new Date(startDateTime)``` to ```moment(startDateTime)```

src/components/FeatureViews/Accounts/TabComponents/InstructorSchedule.js
```diff
@@ -1,132 +1,130 @@
...
+	const teachingSessions = instructor.sessionSet.map(({endDatetime, startDatetime, title}) => ({
+		title: title,
+		start: new Date(startDatetime),
+		end: new Date(endDatetime)
+	}));
+
+	let allDayOOO = false;
+	// Out Of Office
+	const OOOEvents = instructor.instructoroutofofficeSet.map(({startDatetime, endDatetime, description}) => {
+		let start = new Date(startDatetime);
```
CalvinFronda, line 98: Same as above

CalvinFronda (changes requested):
Looks good, just change those new Date to moment

src/components/FeatureViews/Accounts/TabComponents/InstructorSchedule.js
andrewjknapp: I talked with Daniel about this. Turns out FullCalendar requires a plugin to use moment dates, so Daniel said it would be fine to use moment to compare times and use Date objects to use in the FullCalendar events

PR #640, opened 2020-12-01
Updated headings to h1 variant

juicekey:


andrewjknapp (approved):
It looks good. Looks like it is only changing the typography variants like expected.

PR #641, opened 2020-12-01
Dahl design

andrewjknapp:
Upgraded design to Dahl


src/theme/ThemedComponents/Badge/LabelBadge.js
```diff
@@ -0,0 +1,137 @@
+import React from "react";
+import Chip from "@material-ui/core/Chip";
+import theme from "../../muiTheme";
+import { Typography } from "@material-ui/core";
+import PropTypes from "prop-types";
+
+export const LabelBadge = ({label, variant = "default", style, ...rest}) => {
```
dhuangg, line 7: instead of only having a "label" prop, let's also allow the user to submit using a children. Right now to add a label to the LabelBadge component you have to do `<LabelBadge label="something"/>` but it's more natural for developers to use it like `<LabelBadge> something </LabelBadge>` so to do this we need to use react component's children (anything in between an open and close tags). ask @pswk1 if you have questions about this

src/theme/ThemedComponents/Button/ResponsiveButton.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react'
+import { PropTypes } from 'prop-types';
+import Button from "@material-ui/core/Button";
+
+const buttonWidth = (label, children) => {
```
dhuangg, line 5: function names should be verbs. rename to "setButtonWidth"

dhuangg (changes requested):
there's no way I'm going to read through all of these. I'm going to trust most of these are good changes, but I did look at all the new components you all made and made a few comments. Other than that everything looks good to me 👍

src/theme/ThemedComponents/Badge/LabelBadge.js
andrewjknapp: LabelBadges now take label as a child

dhuangg (approved):
Changes look good - ready to merge after Chris's padding update

PR #654, opened 2020-12-09
changed route of add admin action from registration form to admin form

pswk1:
Changed route of add admin action to go to admin form instead of registration form. This solves the issue of having a 400 code rendered on the UI even when admin is successfully created


dhuangg (approved):
yum 🌯

PR #667, opened 2020-12-13
Updated create a class form to reflect availability changes

rgedwill:


dhuangg:
Was able to start fix for #448 

#667 complete but new ticket written for cleaning up the form itself #670

src/components/Form/FormFormats.js
```diff
@@ -178,19 +208,36 @@ export const ACADEMIC_LVL_FIELD = {
     POSITIVE_NUMBER_FIELD = {
         "component": <Fields.TextField />,
         "validator": Yup.number().min(0),
+    },
+    DAY_OF_WEEK_FIELD = {
+
     },
```
NeilAllavarpu, line 214: Either fill this out or remove it

src/components/Form/FormFormats.js
```diff
@@ -1119,46 +1195,114 @@ export default {
...
+                }
+                if (dayAndTime.weekday3 && dayAndTime.startTime3 && dayAndTime.EndTime3) {
+                    availabilities.push(
+                        {
+                            "dayOfWeek": dayAndTime.weekday3,
+                            "startTime": dayAndTime.startTime3 ? dayAndTime.startTime3.format("HH:mm"): null,
+                            "endTime": dayAndTime.endTime3 ? dayAndTime.endTime3.format("HH:mm"): null
+                        }
+                    );
+                }
```
NeilAllavarpu, line 1273: We should try to convert this into a more adaptable method and not just hard-coded to 3 availabilities maximum, probably by using an array or something similar.

NeilAllavarpu (changes requested):
I believe IdleLogout.js has been deleted and migrated to other files on development, check this and migrate your changes accordingly.

src/components/Form/FormFormats.js
dhuangg: good catch but, i'm not spending time on doing it for this PR. pushing this off to #670

PR #668, opened 2020-12-13
Updated dialog title to take h4 stylings

AnnaxGrace:
Updated the muiTheme as well as the individual tags with disableTypography. With research, it looks like this the most recommended way, since dialogTitle (FOR SOME REASON) automatically creates an h2 tag with h6 variant stylings, so overrides and global do not work.

### DIFF
--- a/src/theme/muiTheme.js
+++ b/src/theme/muiTheme.js
@@ -373,10 +373,7 @@ const theme = createMuiTheme({
     //Must use disable typography in the tag on your component
     MuiDialogTitle: {
       root: {
-        fontSize: "16px",
-        fontWeight: "500",
-        fontFamily: "Roboto",
-        lineHeight: "16px",
+          ...h4
       },
     },
     MuiTableCell: {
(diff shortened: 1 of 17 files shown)


src/theme/muiTheme.js
```diff
@@ -36,418 +36,422 @@ export const statusRed = "#FF6766";
...
+        height: "48px",
+      },
+    },
+    //Must use disable typography in the tag on your component
+    MuiDialogTitle: {
+      root: {
+        fontSize: "16px",
+        fontWeight: "500",
+        fontFamily: "Roboto",
+        lineHeight: "16px",
```
dhuangg, line 379: since we have an h4 const at the top, can we do `root: { ...h4 }` instead?

dhuangg (changes requested):
ahh ok so that was what I was missing. I ended up using disableTypography when I was working on this but I had hopes you didn't have to do that for everything for the theme styling to be applied.

src/theme/muiTheme.js
AnnaxGrace: Yes! I had tried to do it like that before and it didn't work... I didn't try using the ... spreader though! It's been updated :)

dhuangg (approved):
10,000 points to youuuu

PR #687, opened 2021-01-20
renamed notes to accountNotes

andrewjknapp:
## Changed notes to account notes in graphql queries to match change in mainframe.


dhuangg (approved):
lovely~

PR #616, opened 2020-11-08
Read More Component

ml5885:
#443 

This is a read more component that adds a read more/ read less link to a text body. You use this component with a self closing tag <ReadMore/>. 

<ReadMore body = {note.body} height = {"20px"}/>

PR #712, opened 2021-01-29
Add item note height

andrewjknapp:
## Goals
- Fix dashboard add button height
- Reimplement ReadMoreText component from PR #616 

### Changes
Made Dashboard addItemButton 50px
Copied ReadMoreText from PR #616 
Added prop to ReadMoreText to pass in custom function to run when Read More is clicked.

### Screenshots
When note is too long for note
image

When Read More is clicked
image


dhuangg (approved):
changes look good to merge - but you mentioned that we should just open up the note to display the entire note which i agree with. I don't think I see this logic here though?

dhuangg (commented):
nevermind i see it. that's smart. haha. Merged!

PR #723, opened 2021-02-11
No results search style

CalvinFronda:


dhuangg (approved):
the no search results page looks great 👍

PR #724, opened 2021-02-11
Unpaid session query 702

### LINKED ISSUE #702: Unpaid Sessions Query - need to migrate to course availabilities
# Description
Course availabilities isn't on the UNAPID_SESSION_QUERY
Screen Shot 2021-01-26 at 3 32 25 PM

Which is causing this error
Screen Shot 2021-01-26 at 3 31 46 PM

CalvinFronda:
Fixes #702


src/components/FeatureViews/AdminPortal/UnpaidSessionCard.js
```diff
@@ -64,15 +64,14 @@ const UnpaidSessionCard = ({ unpaidStudent }) => {
...
-        endTime,
+        courseavailabilitySet,
         hourlyTuition,
     } = unpaidStudent.course;
     const sessionsLeft = unpaidStudent.sessionsLeft;
     const amtDue = adminUtils.amountDue(
         hourlyTuition,
         sessionsLeft,
-        adminUtils.calculateSessionLength(startTime, endTime)
+        adminUtils.calculateSessionLength(courseavailabilitySet.startTime, courseavailabilitySet.endTime)
```
dhuangg: how does this fix the bug? `courseavailabilitySet`is an array so it doesn't have a `startTime` or `endTime` properties. for now, get the first availability's startTime and endTime

dhuangg (changes requested):

src/components/FeatureViews/AdminPortal/UnpaidSessionCard.js
CalvinFronda: You're right, it doesn't fix it. I actually never found the component to trigger the query so i never ran into this error. I made the PR because the change I made didn't effect anything. 
How do you calculate unpaid sessions for sessions that don't have the same time frame?

EX. Instructor is availability 
Mon 9am - 1pm,
Wed 1pm - 3pm,
Fri 8am - 6pm.

Do we need to find the session lengths for each of the instructors availability's?
How would we know what day of week to calculate for the amount due if the instructors schedule is different per week? 
Im not even sure where to get to this on the front-end.

More insight would be nice for this ticket. I know you say for now just use the first index value in the array, but that doesn't solve the issue. I might be overthinking it.

dhuangg: yeah, I've thought of those issues too. This would be a pretty big revamp. The easiest solution would be for backend to calculate the amount due field for us based off of the course's sessions. @jerrylinew @tigermli @NeilAllavarpu

Issue #649, opened 2020-12-08
Scheduler GraphQL Migration


dhuangg:
# Description
The big goal is to migrate the scheduler to use GraphQL. 

# Sessions Query
- timeFrame : DAY, WEEK, MONTH . This filters the sessions to sessions at a day, week, month interval.
- timeShift : integer . Goes up and down a time frame starting from 0. For example, if timeFrame = DAY and timeShift = 0, then you're on today. If timeFrame = Day and timeShift = 1, then you're on tomorrow. If timeFrame = Day and timeShift = -1 then you're on yesterday. 
- viewOption : TUTORING, CLASS, or ALL . This filters sessions by course type.
- startDate, endDate : Gives you all sessions between a start and an end date. 

# Frameworks
- [React Big Calendar](http://jquense.github.io/react-big-calendar/examples/index.html)
- Redux (potentially)
- GraphQL 

# Flow
Request sessions based on UI filters => receive sessions from mainframe => parse session response data to Big Calendar specific session data format => give Big Calendar parsed sessions

Potentially you'll want to store filter values using [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) as well as [React Router's features](https://medium.com/better-programming/using-url-parameters-and-query-strings-with-react-router-fffdcea7a8e9).  

# Features to build
- [x] Fetch all sessions (no filters) from API and display on Calendar View
- [x] List view (toggle between calendar and list views)
- [x] Course, Instructor, Student Session filters 
  - Each of these filters are search selects: you select 1+ items (e.g. course, instructor, student) and filter the displayed sessions matching the filters. We're using react select to filter results on the frontend: https://react-select.com/home
- [x] Pagination between day/week/month. This should flip between days, weeks, months
- [x] Return to "Today" if you're not on the present day/week/month
- [x] Click on session takes you to session view
- [x] (bonus - Calvin will pay you $$$) Session details popover: displays the details of each session when you hover your mouse over 
<img width="879" alt="Screen Shot 2020-12-07 at 4 59 14 PM" src="https://user-images.githubusercontent.com/12959959/101423297-94178880-38ad-11eb-922e-aca1c03e5325.png">

dhuangg:
Need to fix the width of the advance scheduler popover. for some reason it collapses every time i try to set a width

## Extra Long boi
[...]

PR #727, opened 2021-02-12
Scheduler GraphQL Migration

dhuangg:
After much struggle and strife, was able to get basic date filtering and navigation complete. This is to address #649


CalvinFronda (commented):
Logging into a parent account. Click on "Courses" > Click on enrolled course. It will break because 

```
Uncaught TypeError: Cannot read property 'user' of undefined
    at identifyCourseHasEnrolledStudent (CourseClass.js:206)
```

removing ``` student: {user}``` fixes the onCompleted method in the query. 

What is this prop need to do that the global ```user``` variable can't achieve?

src/components/FeatureViews/Scheduler/SessionView.js
```diff
@@ -304,33 +304,24 @@ const SessionView = () => {
...
                 </Grid>
-                <Grid item></Grid>
                 <Grid item>
-                    <Button
-                        className='editButton'
+                    <ResponsiveButton
                         color='primary'
                         onClick={handleEditToggle(true)}
-                        variant='outlined'
+                        variant='contained'
```
CalvinFronda, line 321: A parent shouldn't be able to see the Reschedule button. Suggest adding some account validation.

src/components/Form/FormUtils.js
```diff
@@ -257,6 +257,26 @@ export function arr_diff(a1, a2) {
     return diff;
 }
 
+// Function definition with passing two arrays
```
CalvinFronda, line 260: You can make this faster by using .some & .includes 

```array1.some(item => array2.includes(item))``` 

The issue is ```.includes``` & ```.some``` isn't supported on IE.

**Review by CalvinFronda — CHANGES_REQUESTED** [review]

**Comment by dhuangg** [comment]
> Logging into a parent account. Click on "Courses" > Click on enrolled course. It will break because
> 
> ```
> Uncaught TypeError: Cannot read property 'user' of undefined
>     at identifyCourseHasEnrolledStudent (CourseClass.js:206)
> ```
> 
> removing ` student: {user}` fixes the onCompleted method in the query.
> 
> What is this prop need to do that the global `user` variable can't achieve?

This isn't to compare with the authenticated user (the global user var), this is to see whether or not a parent's student is enrolled in a class. This error may be from the query failing. I'm not getting this error so we'll need to look into this a bit more

File: src/components/Form/FormUtils.js (hunk 2 above)
Comment (dhuangg) [reply src/components/Form/FormUtils.js:260]: i knew there was a smarter way to do this xD. thanks!

File: src/components/FeatureViews/Scheduler/Scheduler.js [hunk 3]
```diff
@@ -41,6 +41,7 @@ const EventPopoverWrapper = ({ children, popover }) => {
     };
 
     const handlePopoverOpen = (event) => {
+        console.log(event.currentTarget);
```
CalvinFronda, line 44: remove log

juicekey (commented):
Looks good to me

PR #732, opened 2021-02-14
removed 2px border caused element to resize on hover

### LINKED ISSUE #710: Weird Contained Button animation bug
# Replication Steps
1. Go to any contained button
2. Hover over it

You will see it flips between having an outline + extra padding and the original contained button version. 

https://user-images.githubusercontent.com/12959959/106305245-49f9b600-6211-11eb-9764-e784dc01ae6f.mov

CalvinFronda:
If you look at the contained button theme we defined border with a 2px as the default.

When adding the style to "&:hover" on the contained property - we never added 2px solid so the hover just adds its own border.

Did we want the border or not?

Fix 1 : remove border in default contained prop.
```  
border: `2px solid ${buttonBlue}`,
```
Fix 2 : Add 2px solid on hover 
```
    border: '2px solid rgba(40, 159, 195, 0.8)',
```

fixes #710

### DIFF
--- a/src/theme/muiTheme.js
+++ b/src/theme/muiTheme.js
@@ -187,7 +187,7 @@ const theme = createMuiTheme({
                 '&:hover': {
                     backgroundColor: buttonBlue,
                     opacity: '80%',
-                    border: 'rgba(40, 159, 195, 0.8)',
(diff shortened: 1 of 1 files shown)


dhuangg (approved):
noice.

Issue #740, opened 2021-03-03
Create invoice row


dhuangg:
# Description
Display the invoice row [More details here](https://docs.google.com/document/d/1aGvBqvhvZSuD0VSuvowiSR6w_hbuE12TyaClO7l5nys/edit#heading=h.9hicnmtnucm9)

# Todo's
- [x] Display the details listed in the product spec for an invoice row.
- [x] Hide the "Customer" column when on parent view

dhuangg:
@CalvinFronda flip the ordering of invoices such that it lists the invoices by largest invoice ID first

Issue #737, opened 2021-03-03
Restyle the Date Range Filter


dhuangg:
# Description 
The date range filter's styling got messed up during the Dahl migration. We should style it back according to the design specifications. 

More details to come after design handoff.

Issue #736, opened 2021-03-03
Invoice Fuzzy Search Component


dhuangg:
# Description
On the invoice view for receptionist, admin, and parent portals, there should be a fuzzy search bar to help with identifying the relevant invoice. Design is still in-progress for this, but we can lay the dev foundation first.

# Todo's
## Receptionist & Admin Portals
- [x] Search invoice by parents or invoice id. Possible inputs: "Kathy" (parent name) or "156" (invoice ID). Possible Options: Parent names and invoice ID's
- [x] The placeholder text should be "Search by Parent or Invoice ID"

## Parent Portal
- [x] Search by invoice ID. Possible inputs "156" Possible Options: Invoice ID's
- [x] The placeholder text should be "Search by Invoice ID"

Issue #735, opened 2021-02-24
[Invoice] Update navigation 


CalvinFronda:
# Update navigation 

### User story 

* PAY-APR-4
* PAY-P-25

## TODO

- [x] Update Navigation: new route to “/invoice” needs to be created that leads to a new Invoices.js view

- [x] Update Navigation bar to say “Invoices” on receptionist, admin, and parent portals. These should link to /invoices => Invoices.js

dhuangg:
pending https://github.com/omou-org/front-end/pull/666 to see if rest of payment routes have been converted to invoice routes

PR #745, opened 2021-03-06
update invoices route

CalvinFronda:
Used ManagePayments file as Invoices placeholder

Payments wern't working so followed the errors and fixed. We should go back and look at renaming most of the variables from Payments -> Invoices to prevent issues in the future.

#735 Invoice route update


CalvinFronda:
fixes #737 
fixes #740 
fixes #736

dhuangg (changes requested):
small syntax changes, but overall looks good!

dhuangg (approved):

Issue #670, opened 2020-12-13
Style Create Course Form


dhuangg:
# Description
The fields aren't in the order or style of the figma
<img width="465" alt="Screen Shot 2020-12-13 at 11 52 02 AM" src="https://user-images.githubusercontent.com/12959959/102022143-b3913400-3d39-11eb-935c-5ba2e2df38d8.png">
<img width="922" alt="Screen Shot 2020-12-13 at 11 52 11 AM" src="https://user-images.githubusercontent.com/12959959/102022144-b55af780-3d39-11eb-8f8f-83d663a4ef4f.png">

# Dev ToDo's
- [ ] Make custom course availabilities form to dynamically accept course availablities
- [ ] Reorder fields to match figma (screenshot above)

dhuangg:
- [ ] update the "Location & Tuition" section to include the fields in [figma](https://www.figma.com/file/YRr3U77NECLQ0EabSr55DC/Omou-Sitemap?node-id=373%3A22745)
- [ ] Style dropdown field for react final form

PR #751, opened 2021-03-17
Fix new course form

andrewjknapp:
# Bring styles on new course form in line with Figma

## Goals

Update page 1 and 3 of the New Course Form to match the order and styling found in the Figma.

## Updates
- Form inputs are placed in the correct order
- Added inputs for Course Link and Google Classroom Class Code
- Updated cache upon course creation for the Registration and Course Management Cache.

**Link to [figma](https://www.figma.com/file/RqyiCN4499Tv1AokoVTqwv/Omou-User-Flows?node-id=395%3A39669)**

### Page 1
image

### Page 3
image

---
#### Note
This PR did not update the styling to Page 2 of the New Course form.

**Related Issue #670**


src/components/Form/FormFormats.js
```diff
@@ -1381,13 +1404,38 @@ export default {
...
+                        name: 'courseLink',
+                        ...stringField(
+                            'Meeting Link',
+                            new_course_form.textFields
+                        ),
+                    },
+                    {
+                        name: 'googleClassCode',
+                        ...stringField(
+                            'GClassromm Enrollment Code',
```
dhuangg, line 1426: Typo here

src/components/Form/FormFormats.js
```diff
@@ -1648,16 +1701,47 @@ export default {
...
+                            const cachedCourseManagementCourses = cache.readQuery(
+                                {
+                                    query: GET_COURSES_BY_ACCOUNT_ID,
+                                    variables: {
+                                        accountId: '',
+                                    },
+                                }
+                            );
+
+                            if (cachedCourseManagementCourses !== null) {
```
dhuangg, line 1731: oh wow. i thought it would usually break on cache.readQuery. this works though?

dhuangg (approved):
1 small typo fix, and we're good to go! I didn't realize that Katie moved the "Back" button on the form to be after the "Cancel" button which feels strange...

src/components/Form/FormFormats.js
andrewjknapp: Yeah this works, if that query does not exist in the cache it returns null, so this handles that case.

PR #755, opened 2021-03-22
Invoice view 747

pswk1:
Couple questions:

- activeAvailabilityList is queried from enrollments field in the Get_Payment query, whereas each course is queried from the enrollment field which is a part of the registrationSet field.
- In the event that an invoice has multiple enrollments in different courses, will the activeAvailabilityList have a way to indicate which availabilities belong to which course? Right now I'm seeing just an arbitrary list of availabilities


dhuangg:
@pswk1 you should be pulling the course availability from the course object so you know which course it belongs to. it's basically a list of times a course meets at (for example a course meets 3 times a week, there will be 3 course availabilities)

pswk1:
An updated preview on invoice detail:

invoice_detail_preview
- Still planning on getting the style details more accurate, but this seems a bit closer to passing the eye test
- Payment status badge, "Days Remaining to Pay", and "Date Due" are hard-coded, these are future to-do's (?)

PR #757, opened 2021-03-28
Invoices style fixes

pswk1:
Invoices styling fix from feedback from designers:
- date range highlight color is omouBlue
- button in date range has no border
- Invoice Table pagination renders conditionally on # of invoices


dhuangg (approved):
these look good!

PR #775, opened 2021-05-17
Fix gc integration mutation

andrewjknapp:
# Fix (GClass Integration) Various Improvements

## Goal 
1. Validate that response from Google contains a valid email for integration
2. Refactor Google Integration mutations
3. Create popup to display when Google response errors out

### 1
Created an **isEmail()** function in utils which uses regex to test if a string is a valid email
Regex pattern taken from [i Hate Regex simple email](https://ihateregex.io/expr/email/)
This is used on the Google response upon integrating to ensure an email was received.
Mutations to consider the user integrated are only called if a valid email was received.

### 2 
Refactored mutations to 
- **INTEGRATE_WITH_GOOGLE** Which takes user id, admin type, and google email and sets those as well as google auth enabled to be true
- **UNINTEGRATE_WITH_GOOGLE** Which takes user id and admin type and then sets google auth enabled to be false

### 3
Created a Dialog on Admin Profile Settings to display when there is an error communicating with Google to get the user's Google account email.
Here is a demonstration of the current popup. The design and contents of the popup are not finalized.

https://user-images.githubusercontent.com/16299570/118562047-d8426b00-b720-11eb-95f9-9514ad64e637.mp4

### DIFF
--- a/src/components/FeatureViews/Accounts/TabComponents/AdminProfileSettings.js
+++ b/src/components/FeatureViews/Accounts/TabComponents/AdminProfileSettings.js
@@ -42,16 +42,15 @@ const ADMIN_GC_ENABLED = gql`
     }
 `;
 
-const SET_ADMIN_GC_ENABLED = gql`
+const UNINTEGRATE_WITH_GOOGLE = gql`
     mutation SetAdminGCEnabled(
         $adminType: AdminTypeEnum!
         $userID: ID!
-        $googleAuthEnabled: Boolean!
     ) {
         createAdmin(
             user: { id: $userID }
             adminType: $adminType
-            googleAuthEnabled: $googleAuthEnabled
+            googleAuthEnabled: false
         ) {
             admin {
                 googleAuthEnabled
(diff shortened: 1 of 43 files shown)


src/components/FeatureViews/Accounts/TabComponents/AdminProfileSettings.js
```diff
@@ -182,41 +183,45 @@ function AdminProfileSettings({ user }) {
...
-            variables: {
-                userID: userInfo.user.id,
-                adminType: userInfo.adminType,
-                googleAuthEmail: response.profileObj.email,
-            },
-        });
+        // Check that Google responded with a valid email
+        if (isEmail(response.profileObj.email)) {
+            setGClassSetting(true);
+            refreshTokenSetup(response).then(() => {
```
dhuangg, line 187: what does "refreshTokenSetup" do? is the idea that if you successfully integrate then you get the courses?

dhuangg (changes requested):
overall looks good - i think i'm still a bit confused by the code in the admin profile settings. maybe renaming some of the variables to explain the logic better will help

src/components/FeatureViews/Accounts/TabComponents/AdminProfileSettings.js
andrewjknapp: Michael found this, it was something to do with making sure we are fetching the newest google classroom token

dhuangg (approved):
you da besttt <3

PR #778, opened 2021-06-04
Admin manage subjects style

pswk1:
- updates to styling of manage course topics tab in admin portal to match Figma specs


src/components/FeatureViews/AdminPortal/CreateTopicModal.js
```diff
@@ -90,57 +95,44 @@ const GET_COURSE_TAGS = gql`
...
-        //                 courseCategory: createdCourseTag
-        //             }
-        //         });
-        // }
+        update: (cache, { data }) => {
+            const newCategory = data.createCourseCategory.courseCategory;
+            const cachedCategory = cache.readQuery({
+                query: GET_COURSE_TAGS,
+            }).courseCategories;
+            const updatedCache = [...cachedCategory, newCategory];
```
dhuangg: Few things:
- Mutations are used for both creating **new** course topics and **updating** existing course topics
- What this means is that this "update" callback function (starting line 108) gets triggered after every creation or update of course topics
- In this function we need to handle both situations (creating a new course topic, or updating an existing course topic)
- Currently the code (line 113) is handling creating new course topics (adding the new category to the end of the list)
- We need to handle the situation when we're updating an existing course topic. To do this, we need to 1) identify where in the course topics (cachedCategory) list is the existing course topic 2) update the course topic there. For example, we want to update the "Math" course topic, so we need to 1) identify where in the list it is (get the index of the "Math" course topic) which we'll say is at index 2. 2) update the course topic at index 2

Your todo will be to implement the logic in the last bullet point

src/components/FeatureViews/AdminPortal/CreateTopicModal.js
```diff
@@ -90,57 +95,44 @@ const GET_COURSE_TAGS = gql`
...
-        //     cache.writeQuery(
-        //         { 
-        //             query: GET_COURSE_TAGS, 
-        //             data: {
-        //                 courseCategory: createdCourseTag
-        //             }
-        //         });
-        // }
+        update: (cache, { data }) => {
+            const newCategory = data.createCourseCategory.courseCategory;
```
dhuangg: refactor to "updatedCourseTopic"

src/components/FeatureViews/AdminPortal/CreateTopicModal.js
```diff
@@ -90,57 +95,44 @@ const GET_COURSE_TAGS = gql`
...
+            const cachedCategory = cache.readQuery({
```
dhuangg: refactor to "cachedCourseTopics"

src/components/FeatureViews/AdminPortal/CreateTopicModal.js
```diff
@@ -90,57 +95,44 @@ const GET_COURSE_TAGS = gql`
     }
 `;
 
-const CreateTagModal = ({ closeModal }) => {
-    const [courseTagData, setCourseTagData] = useState({
-        tagName: '',
-        tagDescription: '',
+const CreateTopicModal = ({ closeModal }) => {
+    const [courseTopicData, setCourseTopicData] = useState({
```
dhuangg: refactor "courseTopicData" to "courseTopicValue". Data = database, Value = state values

dhuangg (changes requested):

src/components/FeatureViews/AdminPortal/CreateTopicModal.js
pswk1: CreateTopicModal.js handles the logic for just creating a new topic (since we're not editing topics inside the modal, right?).
The logic that handles updating the topic at its index is contained in `[ManageCourseTopics](https://github.com/omou-org/front-end/pull/778/commits/d4a3403db8b80ccc0c602b0f0ff493247fdd815e)`, starting at line 139.

pswk1: CreateTopicModal handles the logic of just creating a new topic (since we're not editing topics within this modal, right?)
The logic that implements updating the topic at its current index is contained in ManageCourseTopics.js starting at line 139

dhuangg: ahh gotcha. thanks!

dhuangg (approved):
LOOKS GOOD TO MEEEEE

PR #781, opened 2021-06-16
Fixed Google Classroom call error by re-adding missing function

ml5885:
LoginPage.js was missing the function getCourses(), which was causing an issue since the function was being called.
getCourses() was added back into the file from AdminProfileSettings.js which fixed the issue, with all Google Classroom functionality working.


src/components/Authentication/LoginPage.js
```diff
@@ -175,65 +142,62 @@ const LoginPage = () => {
...
+        refreshTokenSetup(response).then(() => {
+            getCourses();
+        });
     };
-
-    const onFailure = (response) => {
-        console.error(response);
-    };
-
+    const onFailure = () => {};
```
CalvinFronda, line 199: Does this need to do something?

dhuangg (approved):

